PR #1, opened 2024-11-19
initial real plugin

esword:
## Before this PR
See the readme for the "why" of this project

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingExtension.java
```diff
@@ -0,0 +1,39 @@
...
+import org.gradle.api.provider.SetProperty;
+
+public abstract class PluginTestingExtension {
+    public static final String EXTENSION_NAME = "gradleTestUtils";
+
+    /**
+     * Whether to set the ignoreDeprecations system property when running tests.  This is for nebula tests that will
+     * fail if there are gradle deprecations.
+     */
+    public abstract Property<Boolean> getIgnoreGradleDeprecations();
```
esword, line 29: sd

README.md
```diff
@@ -2,11 +2,110 @@
    <a href="https://autorelease.general.dmz.palantir.tech/palantir/gradle-plugin-testing"><img src="https://img.shields.io/badge/Perform%20an-Autorelease-success.svg" alt="Autorelease"></a>
 </p>
 
-## About gradle-plugin-testing
+# About
 
-Gradle plugin and helpers to assist with writing tests for other gradle plugins
+Gradle plugin and helpers to assist with writing tests for other gradle plugins.
```
felixdesouza, line 7: Let's gitve it a proper description, or we get rid of it and rely on the `Primary Functionality` block.

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/TestDepVersions.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Arrays;
+import java.util.Map;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * Utility class to keep versions of dependencies referenced in test files up to date with the versions declared in
+ * the project.
+ */
+public final class TestDepVersions {
```
felixdesouza: can we fully spell out `Dependency`

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/TestDepVersions.java
```diff
@@ -0,0 +1,82 @@
...
+
+        Map<String, String> results = Arrays.stream(
+                        System.getProperty(TEST_DEPENDENCIES_SYSTEM_PROPERTY).split(","))
+                .map(String::trim)
+                .map(dep -> dep.split(":"))
+                .collect(Collectors.toMap(dep -> dep[0] + ":" + dep[1], dep -> dep[2]));
+
+        // TODO(#XXX): do we need to handle just the group?  useful to get anything that is not specifically
+        // referenced from java source sets like the conjure plugin
+        return ImmutableMap.copyOf(results);
```
felixdesouza: why don't we just pass a file as the system property, and you can use normal jackson parsing etc.

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/TestDepVersions.java
```diff
@@ -0,0 +1,82 @@
...
+ * the project.
+ */
+public final class TestDepVersions {
+    static final String TEST_DEPENDENCIES_SYSTEM_PROPERTY = "TEST_DEPENDENCIES";
+    private static final Supplier<Map<String, String>> versionsSupplier =
+            Suppliers.memoize(TestDepVersions::loadVersions);
+
+    /**
+     * Returns the version of the given dependency.  If a specific dependency isn't found, look for a general org
+     * dependency that matches.  Throws exception if neither found.
```
felixdesouza: What is an `org` dependency?

plugin-testing-core/src/test/java/com/palantir/gradle/plugintesting/TestDepVersionsTests.java
```diff
@@ -0,0 +1,46 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class TestDepVersionsTests {
+    private static final String SAMPLE_VERSIONS = "foo:bar:100,com.palantir:gradle-plugin-testing:1.2.3";
+
+    @BeforeAll
+    public static void beforeAll() {
+        System.setProperty(TestDepVersions.TEST_DEPENDENCIES_SYSTEM_PROPERTY, SAMPLE_VERSIONS);
+    }
```
felixdesouza: I was going to say use the `RestorePropertiesExtension` but that's an internal class. You should clear the system property when you're done. 

Similarly, mark `TestDepVersionsTests` as `@Isolated` so it doesn't interfere with other tests.

README.md
```diff
@@ -2,11 +2,110 @@
...
 
-### Built With
+## Using the Plugin
+    
+ ```groovy
+apply plugin: 'com.palantir.gradle.plugin-testing'
```
CRogers: I think `com.palantir.gradle-plugin-testing` reads a bit nicer.

README.md
```diff
@@ -2,11 +2,110 @@
...
+            repositories {
+                mavenCentral()
+            }
+            dependencies {
+                classpath '${resolve('com.palantir.gradle.conjure:gradle-conjure')}'
+            }
+        }
+        dependencies {
+            implementation '${resolve('com.palantir.conjure.java:conjure-lib')}'
+        }
```
CRogers, line 79: Just curious - do you have any examples where we include a non-buildscript dep like this?

README.md
```diff
@@ -2,11 +2,110 @@
...
+class HelloWorldSpec extends IntegrationSpec {
+   def setup() {
+       //language=gradle
+       buildFile << """
+        buildscript {
+            repositories {
+                mavenCentral()
+            }
+            dependencies {
+                classpath '${resolve('com.palantir.gradle.conjure:gradle-conjure')}'
```
CRogers, line 74: The only downside here is (I'm pretty sure) that you get errors when the `//language=gradle` is applied here due to the string interpolation. I wonder if we could just call a method eg:

```groovy
addBuildscriptClasspathDeps(
    'com.palantir.gradle.conjure:gradle-conjure',
    'com.palantir.gradle.jdks:gradle-jdks')
```

that would then add a whole block:

```gradle
buildscript {
    dependencies {
        classpath ...
        classpath ...
    }
}
```

which avoids the red squiggles from interpolation + language insert.

CRogers: Could even do `buildFile << buildscriptClasspathDeps(...)` to keep it flexible for subprojects etc.

plugin-testing-core/src/test/java/com/palantir/gradle/plugintesting/TestDepVersionsTests.java
felixdesouza: Alternatively, you could just write this test in groovy, it'll look slick too, and it has `@RestoreSystemProperties` built in 😎

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersions.java
```diff
@@ -0,0 +1,56 @@
...
+        Set<String> result = new LinkedHashSet<>();
+
+        if (System.getProperty(TEST_GRADLE_VERSIONS_SYSTEM_PROPERTY) == null) {
+            result.addAll(DEFAULT_TEST_GRADLE_VERSIONS);
+        } else {
+            result.addAll(Arrays.asList(
+                    System.getProperty(TEST_GRADLE_VERSIONS_SYSTEM_PROPERTY).split(",")));
+        }
+        return ImmutableList.copyOf(result);
+    }
```
felixdesouza, line 53: I think this should just point to a file as well. Have gradle write the proper thing and then no longer think about it.

README.md
esword: there's some in gradle-baseline and others scrattered about.  Many do not have a version number because they are testing with the GCV plugin, but some do.  Can send you internal search link if you want.

README.md
esword: I figure we'll develop some future fluency calls as we use it more.  You are correct that you get some red squiggles in intellij right now.  I'll look at adding something nicer but would like to get the initial PR through

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -16,13 +16,94 @@
...
+                project.getConfigurations().named(sourceSet.getRuntimeClasspathConfigurationName());
+
+        project.getTasks().withType(Test.class).configureEach(test -> {
+            // need to use the doFirst so that any custom settings on the extension are applied before reading
+            // the values and setting the system properties.
+            Action<Task> action = new Action<>() {
+                @Override
+                public void execute(Task _task) {
+                    // add system property for all test dependencies so that TestDepVersions can resolve them
+                    Set<String> depSet = getDependencyStrings(testRuntimeConfig.get());
```
felixdesouza, line 61: I don't think this necessarily does what you want. I believe you'd also want the integration test source set.

Or basically, whatever source set/classpath is set for this particular `Test` class. We often have "unit" tests and "integration" tests that both run nebula.

README.md
esword: done

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -16,13 +16,94 @@
...
+        addTestDependency(project);
+
+        SourceSetContainer sourceSetContainer = project.getExtensions().getByType(SourceSetContainer.class);
+        SourceSet sourceSet = sourceSetContainer.getByName(SourceSet.TEST_SOURCE_SET_NAME);
+        NamedDomainObjectProvider<Configuration> testRuntimeConfig =
+                project.getConfigurations().named(sourceSet.getRuntimeClasspathConfigurationName());
+
+        project.getTasks().withType(Test.class).configureEach(test -> {
+            // need to use the doFirst so that any custom settings on the extension are applied before reading
+            // the values and setting the system properties.
```
felixdesouza, line 56: I think a clearer distinction would be to have separate tasks for the `TEST_DEPENDENCIES` and `TEST_GRADLE_VERSIONS` blob and feed them as task dependencies.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -16,13 +16,94 @@
...
+
+                    // add system property to ignore gradle deprecations so that nebula tests don't fail
+                    if (testUtilsExt.getIgnoreGradleDeprecations().get()) {
+                        // from
+                        // https://github.com/nebula-plugins/nebula-test/blob/main/src/main/groovy/nebula/test/IntegrationBase.groovy
+                        test.systemProperty("ignoreDeprecations", "true");
+                    }
+                }
+            };
+            test.doFirst(action);
```
felixdesouza, line 78: can you inline the action please 🙏

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
felixdesouza: You can probably go some way further by using the `JvmTestSuite`. Then you can query the test suite, and get its source set and extract the required dependencies. Have a look at: https://github.com/gradle/gradle/blob/master/platforms/jvm/plugins-jvm-test-suite/src/main/java/org/gradle/api/plugins/jvm/JvmTestSuite.java#L57

It does require using https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html but I think that's a worthwhile goal for our gradle plugins and @CRogers moved it over for our internal dev testing plugin. It also makes doing things of this sort a bit easier.

felixdesouza: So the general thing would be to configure each test suite, and then within it configure the `Test` task.

You still can't get around the eager system property thing though, that would have to wait for gradle 9 sadly.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -16,13 +16,94 @@
...
+        });
+    }
+
+    /**
+     * Add test dependency on this plugin to the project so can access utility methods from code if needed.  This is
+     * normally done by getting the Implementation_Version setting in the compiled jar, but that doesn't always work
+     * when running tests so we can also look it up via a gradle property that tests set.
```
felixdesouza: Why doesn't this work?

I feel like we use `Implementation-Version` in a lot of places, I would be more curious about fixing that than adding a separate new property.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -16,13 +16,94 @@
...
+                .or(() -> Optional.ofNullable(
+                        PluginTestingPlugin.class.getPackage().getImplementationVersion()))
+                .orElseThrow(() -> new RuntimeException("PluginTestingPlugin implementation version not found"));
+
+        String testImplConfigName = testSourceSet.getImplementationConfigurationName();
+        project.getConfigurations().named(testImplConfigName).configure(conf -> {
+            conf.getDependencies()
+                    .add(project.getDependencies()
+                            .create("com.palantir.gradle.plugintesting:plugin-testing-core:" + version));
+        });
```
felixdesouza, line 101: this becomes much easier with the `jvm-test-suite` plugin: https://docs.gradle.org/current/userguide/jvm_test_suite_plugin.html#configure_dependencies_of_a_test_suite

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -16,23 +16,210 @@
...
+
+                    then:
+                    println "============std error follows============"
+                    println result.standardError
+                    println "============std out follows============"
+                    println result.standardOutput
+                    result.success
+                }
+                
+                //INSERT MORE TESTS HERE
```
felixdesouza, line 97: I'm not a fan of string replace on big strings, it makes the tests harder to read/understand.

I think it's more than okay to just inline this in the tests themselves

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -16,23 +16,210 @@
...
+            
+            repositories {
+                mavenCentral()
+                mavenLocal()
+            }
+
+            dependencies {
+                implementation gradleApi()
+
+                testImplementation '${resolve("org.junit.jupiter:junit-jupiter")}'
```
felixdesouza, line 48: how are you able to run `resolve` here?

README.md
```diff
@@ -2,11 +2,110 @@
...
+            }
+        }
+        dependencies {
+            implementation '${resolve('com.palantir.conjure.java:conjure-lib')}'
+        }
+        """
+   }
+```
+# Resolution of gradle versions to test against
+Similarly, tests may hardcode versions of gradle that they need to stay compatible with. These versions also get stale and PRs start failing for the inverse reason of the above - the code in the plugin or a dependency of it is updated and is no longer compatible with an old version of gradle. For example, attempting to update jackson libraries from 2.15 -> 2.17 would fail if a test tried to run on gradle < 7.6.4 (when compatibility with jackson 17 was fixed).
```
felixdesouza:
```suggestion
Similarly, tests may hardcode versions of Gradle that they need to stay compatible with. These versions also get stale and PRs start failing for the inverse reason of the above - the code in the plugin or a dependency of it is updated and is no longer compatible with an old version of Gradle. For example, attempting to update jackson libraries from `2.15.0` -> `2.17.0` would fail if a test tried to run on Gradle versiosn < `7.6.4` (when compatibility with jackson `2.17.0` was fixed).
```

README.md
```diff
@@ -2,11 +2,110 @@
...
+
+The `GradleTestVersions` class can provide up-to-date versions of gradle to test against.
```
felixdesouza:
```suggestion
The `GradleTestVersions` class can provide up-to-date versions of Gradle to test against.  
```

README.md
```diff
@@ -2,11 +2,110 @@
...
+    }
+}
+```
+The plugin sets default versions to test against, but these can be overridden using the `gradleTestUtils` extension.  e.g.
+
+```groovy
+gradleTestUtils {
+    gradleVersions = ['7.6.4', '8.8']
+}
+```
```
felixdesouza, line 111: I think you should put this in the context of a wider example e.g. where does this go within your build script.

felixdesouza, line 111: Above comment generally applies. But looking closely, would it not be easier/nicer to have a `test-gradle-versions.props` file that contains the files that we're testing against? It would make excavating changes super easy + more sourcegraph friendly.

Also, I think would we not want to test on whatever version of gradle is running as well? Not blocking however

README.md
```diff
@@ -2,11 +2,110 @@
...
+
+The `GradleTestVersions` class can provide up-to-date versions of gradle to test against.  
+```groovy
+import nebula.test.IntegrationSpec
+import com.palantir.gradle.plugintesting.GradleTestVersions
+
+class HelloWorldSpec extends IntegrationSpec {
+    def 'runs on version of gradle: #version'() {
+        when:
+        gradleVersion = version
```
felixdesouza, line 94: I don't think you want to do it this way. With our internal network, this tries to download from the gradle website and it's excruciatingly slow. It's a lot quicker to actually write a `gradle-wrapper.properties` file with a link to the right thing. This allows you to bypass any sort of mitm slowdown.

README.md
```diff
@@ -2,11 +2,110 @@
...
+
+        then:
+        def result = runTasks('someTask')
+        result.success
+
+        where:
+        version << GradleTestVersions.gradleVersionsForTests
```
felixdesouza, line 101: Not fully developed thought yet, but I'm wondering if we can just make a Spock extension for this. It's also a bit annoying having to add this in every test. You can easily miss it.

Not blocking

README.md
```diff
@@ -2,11 +2,110 @@
...
+```
+
+or in some "Versions" class that has a listing of all the dependencies the test uses.  e.g. 
+```java
+final class TestPluginVersions {
+    static final String CONJURE_JAVA = "com.palantir.conjure.java:conjure-java:5.7.1";
+    static final String CONJURE = "com.palantir.conjure:conjure:4.10.1";
+```
+
+Once these tests are written, the versions of the plugins are often not updated, even when the project under test keeps its dependencies of those plugins up-to-date.  This can cause tests to fail when the plugin is updated not because the plugin is bad, but because there is an incompatibility in the old versions of plugins used in the integration tests.  This can often happen with gradle version bumps.
```
felixdesouza:
```suggestion
Once these tests are written, the versions of the plugins are often not updated, even when the project under test keeps its dependencies of those plugins up-to-date.  This can cause tests to fail when the plugin is updated not because the plugin is bad, but because there is an incompatibility in the old versions of plugins used in the integration tests.  This can often happen with Gradle version bumps.
```

README.md
```diff
@@ -2,11 +2,110 @@
...
+
+When applied to a project, the plugin-testing plugin scans the `testRuntimeClasspath` configuration for the project and passes all dependencies to the test task as a system property.  The version of the dependencies can then be resolved when tests are run and written into generated files.  e.g.
```
felixdesouza:
```suggestion
When applied to a project, the `gradle-plugin-testing` plugin scans the `testRuntimeClasspath` configuration for the project and passes all dependencies to the test task as a system property.  The version of the dependencies can then be resolved when tests are run and written into generated files.  e.g.
```

README.md
felixdesouza: so nebula have an `init.gradle` file that they use as an init script which contain all the dependencies for the buildscript, it's also how you can get your plugin loaded withou having to write a classpath method yourself.

`init.gradle`:
```gradle
allprojects {
    buildscript {
        dependencies {
            classpath files('/Users/fdesouza/.gradle/caches/modules-2/files-2.1/com.fasterxml.jackson.dataformat/jackson-dataformat-xml/2.17.2/ad58f5bd089e743ac6e5999b2d1e3cf8515cea9a/jackson-dataformat-xml-2.17.2.jar')
            classpath files('/Users/fdesouza/.gradle/caches/modules-2/files-2.1/com.palantir.human-readable-types/human-readable-types/1.7.0/c72a0a35db41ce785da5669ed520efd8da944456/human-readable-types-1.7.0.jar')
            // ... more dependencies
        }
    }
}
```

What if we extended said `init.gradle` or added our own, where we set constraints on the versions on the buildscript. Then it means we should be able to just say:

```gradle
buildscript {
  dependencies {
    classpath 'com.palantir.gradle.conjure:gradle-conjure'
  }
}
```

> [!NOTE]
> I haven't actually tried this, I'm just putting together building blocks that feel like they should go together.

I have also been thinking about some sort of gcv for the whole init/settings/buildSrc/buildscript space, and it's somewhat tangentially related to this. Just that the `.props` file is from the project, not the project-under-test.

felixdesouza: I'm also curious with what happens when the main project doesn't actually depend on a dependency, and it's only being included for a particular test. Does this explicitly not cover that case i.e. just write out the version yourself and don't expect it to be up-to-date? If so, it's a bit sad if you do depend on it eventually, you may or may not get the version it seems.

README.md
```diff
@@ -2,11 +2,110 @@
...
+                classpath '${resolve('com.palantir.gradle.conjure:gradle-conjure')}'
+            }
+        }
+        dependencies {
+            implementation '${resolve('com.palantir.conjure.java:conjure-lib')}'
+        }
+        """
+   }
+```
+# Resolution of gradle versions to test against
```
felixdesouza:
```suggestion
# Resolution of Gradle versions to test against
```

README.md
```diff
@@ -2,11 +2,110 @@
...
+This behavior can be overridden using the `gradleTestUtils` extension.  e.g.
+
+```groovy
+gradleTestUtils {
+    ignoreGradleDeprecations = false
+}
+```
+
+## Resolution of dependencies in generated build files
+Integration tests for gradle plugins often write build files which include other plugins.  The version of those included plugins is hardcoded in some way, either directly in the test, e.g.
```
felixdesouza:
```suggestion
Integration tests for Gradle plugins often write build files which include other plugins.  The version of those included plugins is hardcoded in some way, either directly in the test, e.g.
```

README.md
esword: errr, how is this not a proper description?

README.md
esword: cool to know, but I think reasonable for future enhancement to utility for test setup.  This first iteration simply provides the capability to set the versions in existing tests of this format.

README.md
esword: yes, Callum says he hacked some code together that did something like this.  he was going to dig around for it eventually.  but, as you say, not critical for first pass.

README.md
esword: well, the reality is I don't think individual repos _should_ be overriding it because then you are essentially back to the original problem of each repo testing against old versions.  The larger solution is that an organization would write another plugin to set the standard versions that it wants and then apply that plugin to all repos.  This is similar to what we do with latest-jdks, dev-env configs, etc.  Or possibly we change the underlying component to look at a file in the repo with the versions and have an excavator go through and modify that file directly.

We can work out the best approach when we need to start moving versions from where we start them.  This just lays the groundwork for making it possible.

esword: ha.  ya, we're on the same page per my last comment.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
esword: you mean add them as task dependencies to the Test task?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
esword: you mean make it a lambda?  I get an errorprone check about action-lambda-cachable issues then.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
esword: ya - I talked with Callum about the issue of multiple test configurations after the team meeting on Wed.  Our thinking was this will cover the vast majority of gradle plugins because they either have just the standard Test configuration or the testRuntimeClasspath is a base configuration for any other test configuration.  But I agree that if there is a more complete way we can adopt it.  I'm not sure we want to enforce a whole other plugin though.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
esword: that comment is out of date (I moved the required classes into the "core" module) but the functionality may not be.  It came from Callum's work here - https://github.com/palantir/suppressible-error-prone/blob/bf5fe995006346d5d4bf7345a81530f932cc0ca6/gradle-suppressible-error-prone/src/main/java/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePlugin.java#L60-L64

I can re check and see if it's still required in this situation.  Since the use case is more simple it may not be.

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
esword: See the "TODO" above - I hand jam it in for now.  Once we have a published version of the plugin I can remove that System.setProperty and use the published version of the plugin itself

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersions.java
esword: yup - leaning in that direction per above.

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/TestDepVersions.java
esword: done

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/TestDepVersions.java
esword: oh, ya, forgot to update the comment.  Originally when I wrote this it was reading the props and lock files directly, so could have something like "com.palantir.contour:* = x.y.z". The "org" was the group.  But not doing that (at least for now) so will tweak comment

README.md
esword: in that case the author should add a testRuntimeOnly dependency on the version.  That's what these all are - things the tests depend on at runtime.

esword: The init.gradle idea is interesting. I wanted to not be 100% tied to only working in nebula tests since we do have a few repos which use straight gradle toolkit stuff.  But it sounds like a nice enhancement for fluency like Callum's.

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
esword: I am always torn between "Let each test be independent and have all it's own info" and "if you copy common parts you will forget to update something in the future and a test will break in an unexpected way".  Another problem with the former is it's hard to tell what is the unique/interesting part for that test specifically.

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/TestDepVersions.java
esword: hmmm... that would address an issue I fear which is if the list of dependencies is huge we could have a problem with the size of the property value.

but even if I write it I'm not sure I'd need jackson - I think it's going overboard to create some bean with group+name+version as opposed to just using the standard maven syntax "group:name:version"

plugin-testing-core/src/test/java/com/palantir/gradle/plugintesting/TestDepVersionsTests.java
esword: you mean writing as a spock test?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
esword: I did re-check that the version information is not found when running tests without the extra logic of passing the current version as a property.  I even tried changing the `PluginTestingPlugin.class.getPackage().getImplementationVersion()` bit to use the TestDependencyVersions class instead because it is (in theory) loaded from a jar in the tests, but it still fails.  I presume this is because gradle somewhere substitutes a project dependency for the jar dependency.  I updated the comment to be more clear though.

esword:
@CRogers and @felixdesouza - thanks box for the code reviews, especially Felix's which was very (painfully ;-) detailed.  I did the documentation tweaks and couple other things from Felix and the plugin id change that Callum suggested.  Since most of the other suggestions from both of you are essentially internal to how the functionality works (e.g. passing all the settings via files rather than systemProp strings or better ways of setting the classpath dependencies to be more fluent), I propose we get this PR merged so we can have an initial published version of the plugin.  Then we can get an excavator going to start getting it added to all our plugin repos (I'll message you with that PR).  While that is rolling out I can look at some of the enhancements.

The advantage of this is that we can get the automatic functionality of turning on "ignoreDeprecations" in repos that have the plugin applied; No manual changes are required.  This will then allow some number of the repos that are failing the gradle8 upgrade to pass and we can get those off our radar.  We can then more easily see what repos are left (e.g. ones that need the dependencies bumped in tests or that need gradle test versions bumped) and design enhancements to suit them.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
CRogers: When you run a test, it runs using classes for local project files rather than jars. So you can't get the implementation version from the jar, as there is no jar.

esword: yes, I had updated the comment for the method but forgot to push it up.

bmarcaur (approved):

PR #7, opened 2024-11-27
ignore utilities as unused dependencies

esword:
## Before this PR
If an excavator tried to add this plugin to a repository that has the checkUnusedDependencies check enabled, the PR would fail because there the plugin adds a testImplementation dependency on it's own plugin-testing-core module but no code actually references that initially.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


versions.props
```diff
@@ -1,11 +1,13 @@
 com.fasterxml.jackson.core:jackson-databind = 2.18.1
 org.apache.commons:commons-lang3 = 3.17.0
 com.google.guava:guava = 33.3.1-jre
+com.palantir.baseline:gradle-baseline-java = 6.4.0
 
 # test dependencies
 com.netflix.nebula:nebula-test = 10.6.1
 org.assertj:assertj-core = 3.26.3
 org.junit.jupiter:* = 5.11.3
+org.codehaus.groovy:* = 3.0.21
```
esword, line 10: this is something I have seen before in other repos - if there is a dependency on gradle-baseline-java then there must also be a dependency on groovy setting the proper version or else test tasks get an error like "Failed to instantiate spec Conflicting module versions. Module [groovy-xml is loaded in version 3.0.21 and you are trying to load version 3.0.9"

Would be good to figure out how to avoid that, but it's not a problem to be solved in this repo specifically.

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -30,7 +30,7 @@ class PluginTestingPluginIntegrationSpec extends IntegrationSpec {
 
     def setup() {
         //TODO(#xxx): once we have a published version of the plugin, apply it to the project and remove this
-        System.setProperty(TestDependencyVersions.TEST_DEPENDENCIES_SYSTEM_PROPERTY, 'org.junit.jupiter:junit-jupiter:5.11.3,com.netflix.nebula:nebula-test:10.6.1')
+        System.setProperty(TestDependencyVersions.TEST_DEPENDENCIES_SYSTEM_PROPERTY, 'org.junit.jupiter:junit-jupiter:5.11.3,com.netflix.nebula:nebula-test:10.6.1, com.palantir.baseline:gradle-baseline-java:6.4.0')
```
esword, line 33: yes, still need to make it work to apply the plugin to itself.  having a problem with version resolution.  will get there

bmarcaur (approved):

PR #12, opened 2024-12-04
Write dependency versions to file

esword:
## Before this PR
Per @felixdesouza comments on the original PR, passing all the dependency versions as a system property has some downsides.  This PR pulls out that process to a separate task and then makes the test task dependent on it.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -16,25 +16,32 @@
...
 
     def setup() {
-        //TODO(#xxx): once we have a published version of the plugin, apply it to the project and remove this
-        System.setProperty(TestDependencyVersions.TEST_DEPENDENCIES_SYSTEM_PROPERTY, 'org.junit.jupiter:junit-jupiter:5.11.3,com.netflix.nebula:nebula-test:10.6.1, com.palantir.baseline:gradle-baseline-java:6.4.0')
+        writeDependenciesVersionsFile()
 
-        writeHelloWorld('com.testing')
         //language=gradle
         buildFile << """
+            buildscript {
```
esword, line 32: I made the test project use GCV since that more closely resembles real-world projects.  There was an issue where projects with GCV were resolving versions of dependencies differently than I so this plugin wasn't actually working

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/TestDependencyVersionsTask.java
```diff
@@ -0,0 +1,72 @@
...
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Returns a list of all dependencies, sorted and deduplicated.
+     */
+    private static List<String> getDependencyStrings(Configuration config) {
+        // return config.getIncoming().getDependencies().stream()
```
felixdesouza: don't think you need this anymore? the comment that is

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -70,6 +78,7 @@ class PluginTestingPluginIntegrationSpec extends IntegrationSpec {
                             }
                             dependencies {
                                 // This version causes deprecation warnings in gradle 8 for gradle 9
+                                // DO NOT REPLACE THIS VERSION WITH A 'resolve' call
```
felixdesouza: can you explain why?

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -97,13 +106,16 @@ class PluginTestingPluginIntegrationSpec extends IntegrationSpec {
                 //INSERT MORE TESTS HERE
             }
         '''.stripIndent(true)
+
+        TestContentHelpers.addVersionsToPropsFile(file('versions.props'), ['org.junit.jupiter:junit-jupiter', 'com.netflix.nebula:nebula-test', 'com.google.guava:guava', 'com.palantir.gradle.consistentversions:gradle-consistent-versions'])
```
felixdesouza: nit: can you put each dep on a new line 🙏

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/TestDependencyVersionsTaskSpec.groovy
```diff
@@ -0,0 +1,118 @@
...
+            repositories {
+                mavenCentral()
+                mavenLocal()
+            }
+
+            dependencies {
+                implementation 'com.google.guava:guava:33.3.1-jre'
+
+                testImplementation 'org.junit.jupiter:junit-jupiter:5.11.3'
+                testImplementation 'com.netflix.nebula:nebula-test:10.6.1'
```
felixdesouza, line 45: so if I've understood correctly, these versions are static?

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/TestDependencyVersionsTaskSpec.groovy
```diff
@@ -0,0 +1,118 @@
...
+
+    File outputFile
+
+    def setup() {
+        outputFile = new File(projectDir,'build/plugin-testing/dependency-versions.properties')
+    }
+
+    def 'write versions without GCV'() {
+        given:
+        buildFile.text = basicBuildFile
```
felixdesouza: nit: can we just inline the `basicBuildFile` I keep having to scroll up and down 😅

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/TestDependencyVersionsTaskSpec.groovy
```diff
@@ -0,0 +1,118 @@
...
+        TestContentHelpers.addVersionsToPropsFile(file('versions.props'), ['org.junit.jupiter:junit-jupiter', 'com.netflix.nebula:nebula-test', 'com.google.guava:guava', 'com.palantir.gradle.consistentversions:gradle-consistent-versions'])
+        runTasksSuccessfully('writeVersionLocks')
+
+        when:
+        def result = runTasksSuccessfully('writeTestDependencyVersions')
+
+        then:
+        outputFile.exists()
+        !outputFile.text.contains('null')
+        outputFile.text.contains('org.junit.jupiter:junit-jupiter')
```
felixdesouza: can we assert on the version here? not entirely sure what the mechanism is here for why this is working

plugin-testing-core/src/test/java/com/palantir/gradle/plugintesting/TestDependencyVersionsTests.java
```diff
@@ -19,15 +19,30 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 public class TestDependencyVersionsTests {
-    private static final String SAMPLE_VERSIONS = "foo:bar:100,com.palantir:gradle-plugin-testing:1.2.3";
+    private static final String SAMPLE_VERSIONS = "foo:bar=100\ncom.palantir:gradle-plugin-testing=1.2.3";
```
felixdesouza: nit: let's use text blocks!
```suggestion
    private static final String SAMPLE_VERSIONS = """
        foo:bar=100
        com.palantir:gradle-plugin-testing=1.2.3
      """;
```

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/TestDependencyVersionsTaskSpec.groovy
esword: no because the versions will change as the versions within the current project change.  So we could do something else to lookup the versions (i.e. parse the lock file) but that seems a bit strange

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/TestDependencyVersionsTaskSpec.groovy
esword: yes.

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/TestDependencyVersionsTaskSpec.groovy
esword: I couldn't remove the use of TestContentHelpers.addVersionsToPropsFile and just hard-code the set of strings to write into versions.props to make it more clear what we are testing.

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
esword: I revised this to make it easier to understand and not rely on that specific version of the plugin.  I just inserted a task that spit out the string that nebula test looks for.

felixdesouza:
👍

PR #99, opened 2025-05-28
GradleTestVersions should always include `GradleVersion.current()`

crogoz:
## After this PR

GradleTestVersions always includes `GradleVersion.current()`. This will make sure that when gradle-wrapper upgrades, we will test the gradle plugin against the current Gradle Version.

==COMMIT_MSG==
GradleTestVersions should always include `GradleVersion.current()`
==COMMIT_MSG==


felixdesouza (approved):

crogoz:
internal discussion [here](https://pl.ntr/2u4)

PR #111, opened 2025-06-25
Add JUnit5 argument provider for GradleTestVersions

pkoenig10:
This makes it easier for users to create parameterized tests using `GradleTestVersions`.

```diff
  @ParameterizedClass
- @MethodSource("com.palantir.gradle.plugintesting.GradleTestVersions#getGradleVersionsForTests")
+ @GradleTestVersionsSource
  final class MyPluginTest {

      @Parameter
      String gradleVersion

      ...
  }
```


plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsSource.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.junit.jupiter.params.provider.ArgumentsSource;
+
+@Target({ElementType.ANNOTATION_TYPE, ElementType.METHOD, ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@ArgumentsSource(GradleTestVersionsArgumentProvider.class)
```
aldexis, line 31: I don't have strong feelings (if anything, I like the separate class), but wondering whether this _could_ be `@MethodSource("com.palantir.gradle.plugintesting.GradleTestVersions#getGradleVersionsForTests")`?
```suggestion
@MethodSource("com.palantir.gradle.plugintesting.GradleTestVersions#getGradleVersionsForTests")
```

versions.props
```diff
@@ -1,17 +1,12 @@
 com.fasterxml.jackson.core:jackson-databind = 2.18.3
-org.apache.commons:commons-lang3 = 3.17.0
-com.google.guava:guava = 33.4.0-jre
-com.palantir.baseline:gradle-baseline-java = 6.8.0
-
-# test dependencies
```
aldexis, line 6: What led to the refactor of the dependencies here? I guess not all of these were used only in tests?

build.gradle
```diff
@@ -18,7 +18,6 @@ buildscript {
         classpath 'com.palantir.gradle.gitversion:gradle-git-version:3.3.0'
         classpath 'com.palantir.gradle.guide:gradle-guide:1.7.0'
         classpath 'com.palantir.gradle.idea-configuration:gradle-idea-configuration:0.4.0'
-        classpath 'gradle.plugin.org.inferred:gradle-processors:3.7.0'
```
aldexis: nit: was this not used?

aldexis (commented):
Lgtm - Given I don't own this repo though, I'd like a quick double-check from @CRogers though

versions.props
pkoenig10: I find this separation not particularly useful. Where dependencies are used inevitably changes and then this is incorrect. `versions.lock` clearly delineates production and test dependencies.

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsSource.java
pkoenig10: Annotations don't work like this. An annotation isn't like a macro where it is simply expanded to the annotations applied to it - the annotations here apply to the `@GradleTestVersionsSource` macro itself.

build.gradle
pkoenig10: Yes, this plugin has been unnecessary since Gradle 4.6, where Gradle introduced the `annotationProcessor` configuration for better handling for annotation processor dependencies. [The repo](https://github.com/palantir/gradle-processors) has been archived and we have automation that has removed this from all our internal repos (I'll update this automation to target our OSS repos as well).

pkoenig10: Done in https://github.com/palantir/gradle-plugin-testing/pull/115

aldexis (approved):

PR #124, opened 2025-07-09
Don't test Gradle 7 by default

pkoenig10:
Gradle 7 [cannot be used](https://docs.gradle.org/current/userguide/compatibility.html) with tests that are run using Java 21. We are rolling out `runtime = 21` everywhere, so we need to stop testing against Gradle 7.

### DIFF
--- a/plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersions.java
+++ b/plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersions.java
@@ -30,7 +30,7 @@ import java.util.function.Supplier;
  */
 public final class GradleTestVersions {
     static final String TEST_GRADLE_VERSIONS_SYSTEM_PROPERTY = "TEST_GRADLE_VERSIONS";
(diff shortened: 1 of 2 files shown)


plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersions.java
```diff
@@ -30,7 +30,7 @@
  */
 public final class GradleTestVersions {
     static final String TEST_GRADLE_VERSIONS_SYSTEM_PROPERTY = "TEST_GRADLE_VERSIONS";
-    static final List<String> DEFAULT_TEST_GRADLE_VERSIONS = Arrays.asList("7.6.4", "8.8");
+    static final List<String> DEFAULT_TEST_GRADLE_VERSIONS = Arrays.asList("8.14.3");
```
CRogers, line 33: Really, we never should have had default versions like this - I'm not sure how these got here. Regardless, let's update it for now until we get around to fixing this ecosystem.

CRogers (approved):

PR #141, opened 2025-07-23
Add `ConfigurationCacheSpec` as part of gradle-plugin-testing

FinlayRJW:
## Before this PR
This is a FLUP from https://github.com/palantir/gradle-conjure/pull/1637#discussion_r2177218712

Every time we wanted test with configuration cache we would re-create the same `ConfigurationCacheSpec`

## After this PR
Now we have a centralised place for `ConfigurationCacheSpec` so it can be re-used 

==COMMIT_MSG==
Add `ConfigurationCacheSpec` as part of gradle-plugin-testing
==COMMIT_MSG==


gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -50,7 +50,8 @@ class PluginTestingPluginIntegrationSpec extends AbstractTestingPluginSpec {
             dependencies {
                 implementation gradleApi()
                 implementation 'com.google.guava:guava'
-
+                
+                testImplementation 'org.codehaus.groovy:groovy'
```
FinlayRJW, line 54: I had to add this otherwise a bunch of tests would fail with :

```
* What went wrong:
Execution failed for task ':writeTestDependencyVersions'.
> Could not resolve all files for configuration ':testRuntimeClasspath'.
   > Could not resolve org.codehaus.groovy:groovy:{strictly 3.0.12}.
     Required by:
         project :
      > Cannot find a version of 'org.codehaus.groovy:groovy' that satisfies the version constraints:
           Dependency path ':ignoreDeprecations-automatically-set-when-plugin-applied:unspecified' --> 'com.palantir.gradle.plugintesting:configuration-cache-spec:0.7.0-9-g44ccdc9.dirty' (runtime) --> 'org.codehaus.groovy:groovy:3.0.21'
           Constraint path ':ignoreDeprecations-automatically-set-when-plugin-applied:unspecified' --> ':ignoreDeprecations-automatically-set-when-plugin-applied:unspecified' (gcvLocks) --> 'org.codehaus.groovy:groovy:{strictly 3.0.12}' because of the following reason: Locked by versions.lock
           Dependency path ':ignoreDeprecations-automatically-set-when-plugin-applied:unspecified' --> 'com.palantir.gradle.plugintesting:configuration-cache-spec:0.7.0-9-g44ccdc9.dirty' (runtime) --> 'org.spockframework:spock-core:2.3-groovy-3.0' (runtimeElements) --> 'org.codehaus.groovy:groovy:3.0.12'
           Dependency path ':ignoreDeprecations-automatically-set-when-plugin-applied:unspecified' --> 'com.netflix.nebula:nebula-test:10.6.2' (runtimeElements) --> 'org.spockframework:spock-junit4:2.3-groovy-3.0' (runtimeElements) --> 'org.codehaus.groovy:groovy:3.0.12'
```

when I tested this locally in another repo though I didn't need to add this so think its related to the test setup?

configuration-cache-spec/src/main/groovy/com/palantir/gradle/plugintesting/ConfigurationCacheSpec.groovy
```diff
@@ -0,0 +1,73 @@
...
+import org.gradle.testkit.runner.BuildResult
+
+
+abstract class ConfigurationCacheSpec extends IntegrationTestKitSpec {
+    def setup() {
+        definePluginOutsideOfPluginBlock = true
+        keepFiles = true
+    }
+
+    // keep parity with old fileExists method
```
kelvinou01: Is this needed here? Let's have the subclasses of `ConfigurationCacheSpec` define it themselves?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -102,7 +106,7 @@ public void execute(Task _task) {
     }
 
     /**
-     * Add test dependency on the utility jar to the project so that an explicit dependency statement isn't needed.
+     * Add test dependency on the utility jars to the project so that an explicit dependency statement isn't needed.
```
kelvinou01, line 109: On a broader note, I'm not sure how I feel about implicitly adding test dependencies like this. 

It's already hard enough to answer "where is this coming from???" (where **this** can be a task, configuration, dependency, etc) in Gradle. It feels like `addTestDependency` is adding to the slog. 

Maybe we can leverage jvm-test-suites as suggested by @felixdesouza [here](https://github.com/palantir/gradle-plugin-testing/pull/1/files#r1852638732), as a follow up?

kelvinou01 (commented):
Nice PR! Let's follow up this by replacing all the individual ConfigurationCache*.groovy definitions scattered everywhere with this!

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
FinlayRJW: yeah sounds like a good idea, feels separate from what we are doing here and more of a general improvement to `gradle-plugin-testing` so will note as a FLUP

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
esword: We want the test setup to match a realistic project setup.  I'd dig a bit more to figure out what changed that requires this to be added.  (Yes, I realize it's not fun.)

FinlayRJW: okay so I think the issue here is that we run setup which adds all the dependencies and versions.props then we write locks - but then later in a test we add the plugin via `applyTestUtilsPlugin` now as my changes bumps the minimum version of `org.codehaus.groovy:groovy` and this causes the `writeTestDependencyVersions`  error

To resolve this all you need todo is run `wVL` after you add the plugin or do what I've done here and pin it at the higher version. 

I guess the question is wether its acceptable that a user would need to re-run `wVL` I think this should be okay as the excavator already runs:

```java
    @Override
    protected void pluginUpdated(Optional<String> fromVersion, OrderableSlsVersion toVersion) {
        signalWriteLocks();
    }
```

esword (approved):

PR #143, opened 2025-07-24
make sure works with older versions as needed

esword:
## Before this PR
Tests did not use the plugins own GradleTestVersions.gradleVersionsForTests

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


FinlayRJW:
👍

PR #152, opened 2025-07-31
Make `gradle-plugin-testing` compatible with the Configuration Cache

kelvinou01:
## Before this PR

We are rolling out the Configuration Cache (CC). To start, we are trying to get `gw classes` and `gw test` to work with the configuration cache. To achieve that, `PluginTestingPlugin` has to be made CC compatible by solving the following CC violations:

- `Test` was capturing `TestDependencyVersionsTask` as an input
- `TestDependencyVersionsTask` was taking in `Configuration` as an input

## After this PR

Wire up `Test` and `TestDependencyVersionsTask` with providers, and narrow the input type of `TestDependencyVersionsTask`.

==COMMIT_MSG==
Make `gradle-plugin-testing` compatible with the Configuration-Cache
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
+++ b/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
@@ -98,27 +98,32 @@ public class PluginTestingPlugin implements Plugin<Project> {
         project.getTasks().withType(Test.class).configureEach(test -> {
             test.dependsOn(testDependencyVersions);
 
-            // need to use the doFirst so that any custom settings on the extension are applied before reading
-            // the values and setting the system properties.
-            Action<Task> action = _task -> {
-                // add system property for name of file to read for dependency versions
-                test.systemProperty(
-                        TestDependencyVersions.TEST_DEPENDENCIES_FILE_SYSTEM_PROPERTY,
-                        testDependenciesFileAbsolutePath.get());
+            // doFirst so any custom settings on the extension are applied before reading the values and setting
+            // the system properties.
 
-                // add system property for what versions of gradle should be used in tests
-                String versions =
-                        String.join(",", testUtilsExt.getGradleVersions().get());
-                test.systemProperty(GradleTestVersions.TEST_GRADLE_VERSIONS_SYSTEM_PROPERTY, versions);
+            // Do not replace with a lambda! Lambdas aren't build cacheable.
+            // https://github.com/gradle/gradle/issues/5510
+            test.doFirst(new Action<>() {
+                @Override
+                public void execute(Task _task) {
+                    // add system property for name of file to read for dependency versions
+                    test.systemProperty(
+                            TestDependencyVersions.TEST_DEPENDENCIES_FILE_SYSTEM_PROPERTY,
+                            testDependenciesFileAbsolutePath.get());
 
-                // add system property to ignore gradle deprecations so that nebula tests don't fail
-                if (testUtilsExt.getIgnoreGradleDeprecations().get()) {
-                    // from
-                    // https://github.com/nebula-plugins/nebula-test/blob/main/src/main/groovy/nebula/test/IntegrationBase.groovy
-                    test.systemProperty("ignoreDeprecations", "true");
+                    // add system property for what versions of gradle should be used in tests
+                    String versions =
+                            String.join(",", testUtilsExt.getGradleVersions().get());
+                    test.systemProperty(GradleTestVersions.TEST_GRADLE_VERSIONS_SYSTEM_PROPERTY, versions);
+
+                    // add system property to ignore gradle deprecations so that nebula tests don't fail
+                    if (testUtilsExt.getIgnoreGradleDeprecations().get()) {
+                        // from
+                        // https://github.com/nebula-plugins/nebula-test/blob/main/src/main/groovy/nebula/test/IntegrationBase.groovy
+                        test.systemProperty("ignoreDeprecations", "true");
+                    }
                 }
-            };
-            test.doFirst(action);
+            });
         });
     }
 
--- a/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
+++ b/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
@@ -16,6 +16,7 @@
 package com.palantir.gradle.plugintesting;
 
 import com.palantir.baseline.tasks.CheckUnusedDependenciesParentTask;
+import com.palantir.gradle.plugintesting.TestDependencyVersionsTask.TestDependency;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -25,8 +26,6 @@ import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
-import org.gradle.api.artifacts.ModuleVersionIdentifier;
-import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.provider.Provider;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.SourceSetContainer;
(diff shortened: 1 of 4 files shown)


gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -67,38 +75,47 @@ private static void doApply(Project project) {
...
 
+        Provider<String> testDependenciesFileAbsolutePath = project.provider(() ->
+                testDependencyVersions.get().getOutputFile().get().getAsFile().getAbsolutePath());
+
         project.getTasks().withType(Test.class).configureEach(test -> {
             test.dependsOn(testDependencyVersions);
 
             // need to use the doFirst so that any custom settings on the extension are applied before reading
             // the values and setting the system properties.
-            Action<Task> action = new Action<>() {
```
kelvinou01, line 78: Intellij nagged me to make this a lambda ><

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -67,38 +75,47 @@ private static void doApply(Project project) {
...
-                @Override
-                public void execute(Task _task) {
-                    // add system property for name of file to read for dependency versions
-                    test.systemProperty(
-                            TestDependencyVersions.TEST_DEPENDENCIES_FILE_SYSTEM_PROPERTY,
-                            testDependencyVersions
```
kelvinou01, line 84: Removed the capture on the `testDependencyVersions` task here

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
FinlayRJW: I don't think we want to convert this to a lambda - [see this errorprone](https://github.com/palantir/gradle-baseline/blob/develop/baseline-error-prone/src/main/java/com/palantir/baseline/errorprone/GradleCacheableTaskAction.java) 

I think we're not catching this here as the action is not inlined if it is we see:

```
> Compilation failed; see the compiler output below.
  /Volumes/git/gradle-plugin-testing/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java:76: error: [GradleCacheableTaskAction] Gradle task actions are not cacheable when implemented by lambdas
              test.doFirst(_task -> {
                           ^
      (see https://github.com/palantir/gradle-baseline#baseline-error-prone-checks)
  1 error

```

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/TestDependencyVersionsTask.java
```diff
@@ -21,37 +21,35 @@
...
 public abstract class TestDependencyVersionsTask extends DefaultTask {
+    public static final String FILENAME = "plugin-testing/dependency-versions.properties";
 
     public TestDependencyVersionsTask() {
-        getOutputFile()
-                .convention(getProject()
-                        .getLayout()
-                        .getBuildDirectory()
-                        .file("plugin-testing/dependency-versions.properties"));
+        getOutputFile().convention(getProject().getLayout().getBuildDirectory().file(FILENAME));
```
FinlayRJW: can we remove the use of `@getProject` while we are here?

```java
    @Inject
    protected abstract ProjectLayout getProjectLayout();
```

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
CRogers: It bugs me that this `Action` is assigned to a variable rather than just inline with `doFirst` - I have a feeling our errorprones dealing with task actions won't work unless it's inlined right?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -67,38 +75,47 @@ private static void doApply(Project project) {
...
                     NamedDomainObjectProvider<Configuration> testRuntimeConfig =
                             project.getConfigurations().named(sourceSet.getRuntimeClasspathConfigurationName());
-                    task.getClasspathConfiguration().set(testRuntimeConfig);
+                    Provider<Iterable<ModuleVersionIdentifier>> testRuntimeDependencies = testRuntimeConfig.map(
+                            config -> config.getIncoming().getResolutionResult().getAllComponents().stream()
+                                    .map(component -> {
+                                        ModuleVersionIdentifier moduleInfo = component.getModuleVersion();
+                                        if (moduleInfo == null) {
+                                            log.info(
+                                                    "Component with null module version will not be included in {}: {}",
```
CRogers: Does this happen often? Can you give an example?

kelvinou01: Not sure how often this happens, but [`getModuleVersion()` is a @Nullable method in the `ResolvedComponentResult` interface](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/result/ResolvedComponentResult.html#getModuleVersion()), so I'm checking and logging here just to be sure. 

I didn't want any nullpointers to cause test failures for consumers.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/TestDependencyVersionsTask.java
kelvinou01: I think it should be fine to keep `getProject`s during the configuration phase ><

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
kelvinou01: Good catch @FinlayRJW ! I'll inline the action and revert it to an anonymous class.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/TestDependencyVersionsTask.java
kelvinou01: Let's do it anyways for aesthetic reasons

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
CRogers: I think once resolution has happened every module has to have a version. If a version can't be found, the `getResolutionResult` method throws. So you can just not check for nulls here. We don't anywhere else we do this (which is 10s of places).

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -67,26 +71,31 @@ private static void doApply(Project project) {
                     SourceSet sourceSet = sourceSetContainer.getByName(SourceSet.TEST_SOURCE_SET_NAME);
                     NamedDomainObjectProvider<Configuration> testRuntimeConfig =
                             project.getConfigurations().named(sourceSet.getRuntimeClasspathConfigurationName());
-                    task.getClasspathConfiguration().set(testRuntimeConfig);
+                    Provider<Iterable<ModuleVersionIdentifier>> testRuntimeDependencies = testRuntimeConfig.map(
+                            config -> config.getIncoming().getResolutionResult().getAllComponents().stream()
+                                    .map(ResolvedComponentResult::getModuleVersion)
+                                    .collect(Collectors.toUnmodifiableSet()));
```
CRogers: So before, this did `config.getResolvedConfiguration().getFirstLevelModuleDependencies()` and now we're getting all transitives by using `getAllComponents` - I think we want to keep the old behaviour, as in this plugin we only care about the dependencies that are explicitly listed like so and not the transitives they bring in:

```
dependencies {
    testRuntimeOnly 'something'
}
```

kelvinou01: I originally switched from `getResolvedConfiguration()` to `getIncoming()` because the former was being deprecated. However, seems like there's no easy way to get just the first level deps from getResolvedConfiguration. 

I checked and `getResolvedConfiguration()` is still in Gradle 9. Let's stick to that for now to keep functionality parity, since we're going to refactor this repo soon anyways.

kelvinou01:
Also just noting down/flagging that on [eaddb6e](https://github.com/palantir/gradle-plugin-testing/pull/152/commits/eaddb6e83b840b4fdb95c7af34e897817e91d63d) I'm getting this error locally, where the baseline implicit dependency checker thinks something should depend on itself.

```
Found 1 implicit dependencies - consider adding the following explicit dependencies to 'configuration-cache-spec/build.gradle', or avoid using classes from these jars:
    dependencies {
        implementation 'com.palantir.gradle.plugintesting:configuration-cache-spec'
    }

```

CRogers (approved):

PR #196, opened 2025-09-30
Transitively bring in nebula-test for consumers by making it a `api` dependency

kelvinou01:
## Before this PR

Consumers of configuration-cache-spec have to bring in nebula-test explicitly for their ConfigurationCacheSpecs to be recognized as tests. 

## After this PR

==COMMIT_MSG==
Transitively bring in nebula-test for consumers by making it a `api` dependency
==COMMIT_MSG==


FinlayRJW:
👍

PR #209, opened 2025-10-10
New JUnit based testing framework for Gradle Plugins

CRogers:
## Before this PR
We currently do our Gradle plugin using Groovy based tests, using a variety of different frameworks:
1. nebula-test `IntegrationSpec`: The most common. Does not work with configuration cache. Great debugging experience.
2. nebula-test `IntegrationTestKitSpec`: Like `IntegrationSpec` (but slightly different) - work with configuration cache.
3. This repo's `ConfigurationCacheSpec`: A wrapper around `IntegrationTestKitSpec` with some extra helpers for testing that configuration cache works.
4. Using gradle-test-kit directly: usually with some adhoc abstractions (eg in gradle-baseline). Poor debugging experience.
5. Internally, some people using gradle-test-kit directly but using Java.

There are problems with these:
1. They all use Groovy
   1. Every time Gradle upgrades Groovy (such as the upcoming Gradle 9) we are in a world of pain. Groovy has a runtime component that much match up _exactly_ with the version used to compile it. This issue rears it's head _all the time_ when trying to test two Gradle versions with different Groovy major versions from the same test task:
       > groovy.lang.GroovyRuntimeException: Conflicting module versions. Module [groovy-xml is loaded in version 3.0.24 and you are trying to load version 3.0.25
   2. We are primarily a Java company, and virtually know one who is not on the developer tooling team understands groovy to any real level beyond using some Gradle buildscripts.
   3. As unknown as Groovy is, Spock is even more unknown
   4. We have very little Groovy infrastucture compare to our extensive Java infrastructure (esp errorprone, formatter, other static analysis)
2. Such a wide variety of different testing approaches is confusing.

## After this PR
==COMMIT_MSG==
Introduce a new Java based JUnit extension to help easily test Gradle plugin code
==COMMIT_MSG==

Features:

### It's written in Java

* No Groovy module mismatches every again! Gradle 9 is easy to test against.
* Our devs already use Java and JUnit, very easy to pick up.
* Infrastructure like errorprones, formatter etc just work!

### Runs against multiple versions of Gradle at once with no boilerplate

Uses a new feature of JUnit called `ClassTemplateInvocationContextProvider`, which allows us to run all the tests in a class multiple times with different versions of Gradle:

Screenshot 2025-10-10 at 16 33 57

This allows us to automatically change the versions of Gradle we test against in the repo, ie keep repos up-to-date with testing against the latest Gradle release and latest minor release of each major version number.

### Automatic language injection into intellij

* No need for manual `// language=Gradle` or `// language=Java` strewn everywhere!
* You get syntax highlight/code completion/navigation for Gradle groovy buildscripts, Java, yaml, properties files - we can add more later!

Screenshot 2025-10-10 at 16 24 37

### Nice structured API for actually creating files in test setups

* eg the `rootProject.mainSourceSet().java().writeClass` you saw above
* Easily make subprojects by injecting `SubProject subProjectName` into you test method

### Use gradle-test-kit, but completely wraps the implementation

We get all the benefits of using test kit, but we wrap the implementation completely which gives us a few benefits:
1. We can easily hang extra behaviour we like on the wrapper classes
2. Handle breaks or changes in test kit at the library level rather than in each repo
3. Change our how execute Gradle (eg some tests may actually want to call `./gradlew` directly, eg tests in `gradle-jdks`.

## Possible downsides?
We [risk of adding one more way of testing](https://xkcd.com/927/) and not following through and changing the existing tests.


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvocation.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.testing.execution;
+
+import java.util.Map;
+import org.gradle.testkit.runner.GradleRunner;
+
+public final class GradleInvocation {
+    private final GradleRunner gradleRunner;
```
CRogers, line 23: If you're wondering why we're wrapping all the test kit types, read [here](https://github.com/palantir/gradle-plugin-testing/pull/209/files#diff-36783d669b50c2e89eb6f52fd8354063b2bb0db9878546e86d5bf9ccd8b318abR15-R20)

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java
```diff
@@ -0,0 +1,183 @@
...
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
+import com.palantir.gradle.testing.execution.GradleInvocation;
+import com.palantir.gradle.testing.execution.GradleInvoker;
+import com.palantir.gradle.testing.execution.InvocationResult;
+import com.palantir.gradle.testing.execution.TaskOutcome;
+import com.palantir.gradle.testing.junit.GradlePluginTests;
+import com.palantir.gradle.testing.project.RootProject;
+import com.palantir.gradle.testing.project.SubProject;
+import org.junit.jupiter.api.Test;
```
CRogers, line 29: I would start reading this PR from these "usages" tests onward, to get a sense of what the tests look like.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/properties/PropertiesFile.java
```diff
@@ -0,0 +1,42 @@
...
+import java.nio.file.Path;
+import org.intellij.lang.annotations.Language;
+
+public record PropertiesFile(Path path) implements ProjectFile<PropertiesFile> {
+    public PropertiesFile appendProperty(String key, String value) {
+        return appendLine("%s=%s".formatted(key, value));
+    }
+
+    @Override
+    public PropertiesFile overwrite(@Language("Properties") String text) {
```
felixdesouza, line 34: I imagine these are just overridden so you can add the `@Language(...)` blocks?

CRogers: yep - that's pretty much the whole reason. But the class also gives you a nice place to add more behaviour - see the `addProperty` method above we use in a test.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
```diff
@@ -0,0 +1,81 @@
...
+
+    default T edit(UnaryOperator<String> editor) {
+        return overwrite(editor.apply(text()));
+    }
+
+    default T createEmpty() {
+        return overwrite("");
+    }
+
+    default AbstractPathAssert<?> assertThat() {
```
felixdesouza, line 70: this is nice

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
```diff
@@ -0,0 +1,81 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardOpenOption;
+import java.util.Arrays;
+import java.util.function.UnaryOperator;
+import java.util.stream.Stream;
+import org.assertj.core.api.AbstractPathAssert;
+import org.assertj.core.api.Assertions;
+
+public interface ProjectFile<T extends ProjectFile<T>> {
```
felixdesouza, line 30: this is what allows you to return "this" and it keeps the correct types or?

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersions.java
```diff
@@ -29,7 +29,7 @@
  * plugin is both backwards and forwards compatible.
  */
 public final class GradleTestVersions {
-    static final String TEST_GRADLE_VERSIONS_SYSTEM_PROPERTY = "TEST_GRADLE_VERSIONS";
+    static final String TEST_GRADLE_VERSIONS_SYSTEM_PROPERTY = "com.palantir.gradle.testing.gradle_versions_to_test";
```
felixdesouza, line 32: intentional change?  if so, I do like it!

nit: can we just have snake case? don't care too much if you care more

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -0,0 +1,73 @@
...
+        rootProject
+                .buildGradle()
+                .append(
+                        """
+            println "hello from ${path}"
+            println "project name: ${name}"
+            """);
+
+        assertThat(gradle.withArgs().buildsSuccessfully().output()).contains("hello from :");
+        assertThat(gradle.withArgs().buildsSuccessfully().output()).contains("project name: root");
```
CRogers, line 37: So currently this lets people name subprojects using the parameter name ie you inject a `SubProject foobar` and you get a subproject called `foobar`. However, this is not the case for root projects. It's currently fixed to always be `root`. This is kinda purposefully in contrast to `IntegrationSpec` where it gets the name of test dir, which is kinda nasty/annoying.

But my question is, should be take the root project name from the `RootProject blah` injected parameter? We definitely do set the root project name in some existing tests, but I think that may be because the default `IntegrationSpec` behaviour is so nasty, rather than it actually being necessary to be something special.

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/GradleVersionsFromJunitParameterFixtureTest.java
```diff
@@ -0,0 +1,39 @@
...
+    void testName(GradleInvoker gradleInvoker, RootProject rootProject) {
+        rootProject
+                .buildFile()
+                .append(
+                        """
+            import org.gradle.util.GradleVersion
+            println "GradleVersion: ${GradleVersion.current().version}"
+            """);
+
+        throw new RuntimeException(gradleInvoker.withArgs().buildsSuccessfully().output());
```
felixdesouza, line 36: this is unfinished?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
CRogers: Yeah, it's the classic ["curiously recurring template pattern"](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern)

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/GradleVersionsFromJunitParameterFixtureTest.java
CRogers: No, this is a test fixture, look at `GradleVersionsFromJunitParameterTest`

CRogers: I'll add a comment

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public GradleInvocation withArgs(String... args) {
+        return new GradleInvocation(GradleRunner.create()
+                .withProjectDir(rootProjectDir.toFile())
+                .withDebug(false)
+                .forwardOutput()
+                .withGradleVersion(gradleVersion.version())
+                .withPluginClasspath()
+                .withArguments(args));
```
FinlayRJW, line 41: do we want to add some default args i.e. run with configuration cache? so the user has to explicitly add `--no-configuration-cache`?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/java/JavaSrcDirTest.java
```diff
@@ -0,0 +1,169 @@
...
+    Path srcDirPath;
+
+    JavaSrcDir javaSrcDir;
+
+    @BeforeEach
+    void beforeEach() {
+        javaSrcDir = new JavaSrcDir(srcDirPath);
+    }
+
+    @Nested
```
felixdesouza, line 38: Definitely think we should make use of `@Nested` more! Neat

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/java/JavaSrcDirTest.java
```diff
@@ -0,0 +1,169 @@
...
+        void writes_interface_out_to_correct_location() {
+            String javaSource =
+                    """
+                package foo;
+                interface SomeInterface {}
+                """;
+
+            javaSrcDir.writeClass(javaSource);
+
+            assertThat(srcDirPath.resolve("foo/SomeInterface.java")).hasContent(javaSource);
```
felixdesouza, line 61: will need to read code a bit more, but why this matter per kind? is this just a parsing verification  thing?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
CRogers: I think that's definitely a feature we want to add in a PR after this merges. But we need to think about the rollout. I think ideally we'd want to do something where we add a boolean property in the gradle plugin extension like `testWithConfigurationCacheByDefault`. Then perhaps an annotation on tests or test classes that is like `@NotCompatibleWithConfigurationCache("Automatically disabled by excavator for rollout")` which does not turn on CC for those tests. We then excavate out the property, see which tests fail, mark them with the annotation, and roll out without stalling upgrades. Also then have a burndown list of configuration cache issues to fix via sourcegraph!

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/java/JavaSrcDirTest.java
CRogers: These are just unit tests for the parsing to check it works for all of classes/interfaces/records etc

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersions.java
CRogers: The "style" for junit configuration parameters is to prefix with a namespace like `com.palantir.gradle.testing.` or `junit.jupiter.`, so I'd like to keep the dots for that part at least.

This is all internal, so shouldn't cause any breaks unless someone has managed to get different versions of the plugin vs the testing library, but this would be hard to do as the plugin sets up all the deps.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
CRogers: yeah, I think there are more places we can put stuff like this. Eg on the `InvocationResult` class and stuff it references.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
CRogers: This assumes we have migrated all the test to this new framework.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java
```diff
@@ -0,0 +1,183 @@
...
+                // Needs a task action to not be up-to-date
+                doLast {}
+            }
+            """);
+
+        InvocationResult result = gradle.withArgs("foo").buildsSuccessfully();
+
+        assertThat(result.task(":subProject:foo")).hasValueSatisfying(taskResult -> {
+            assertThat(taskResult.outcome()).isEqualTo(TaskOutcome.SUCCESS);
+        });
```
felixdesouza: obviously not for this PR, but I think it would be nice to add our own assertions, akin to the `SafeLoggableAssertions`, for example:

```java
assertThat(result)
  .ranTask(":subproject:foo")
  .withOutcome(TaskOutcome.SUCCESS);
```

Without it, there's just a lot of machinery to check that a task run is successful

felixdesouza: note, this doesn't preclude you from writing this way at all! IIUC, you'd be extending some `Assertion` class. This is what safe-logging does:

https://github.com/palantir/safe-logging/blob/c6bbe2ca6eaa9f345936fbb7ce26f92be6e3dee2/preconditions-assertj/src/main/java/com/palantir/logsafe/testing/LoggableExceptionAssert.java#L28-29

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java
```diff
@@ -0,0 +1,183 @@
...
+        assertThat(result.task(":i-dont-exist")).isEmpty();
+    }
+
+    @Test
+    void excluded_tasks_are_returned_as_optional_empty(GradleInvoker gradle, RootProject rootProject) {
+        rootProject.buildFile().appendLine("tasks.register('foo')");
+
+        InvocationResult result = gradle.withArgs("foo", "-x", "foo").buildsSuccessfully();
+
+        assertThat(result.task(":foo")).isEmpty();
```
felixdesouza, line 75: this could be:

```java
assertThat(result)
  .didNotRunTask(":foo")
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradlePluginTests.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.junit.jupiter.api.ClassTemplate;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(GradleVersioningClassTemplate.class)
+@ClassTemplate
```
felixdesouza, line 30: I wonder whether you can use parameterized classes as well as class templates i.e. have the tree:

```
Gradle Version 8.14.3
  - paramter 1 value 1
  - parameter 1 value 2
Gradle Version 9.1.0
  - parameter 2 value 1
  - parameter 2 value 2
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleVersioningClassTemplate.java
```diff
@@ -0,0 +1,64 @@
...
+        @Override
+        public void prepareInvocation(ExtensionContext context) {
+            GradleVersionStore.setGradleVersion(context, gradleVersion);
+        }
+
+        @Override
+        public List<Extension> getAdditionalExtensions() {
+            return List.of(new GradleInvokerParameterResolver(), new GradleProjectParameterResolver());
+        }
+    }
```
felixdesouza, line 63: this is very neat!

felixdesouza: fortuitous that this was released relatively recently (30th May 2025)

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/GradleVersionsFromJunitParameterFixtureTest.java
felixdesouza: oh this is a test fixture

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java
```diff
@@ -0,0 +1,183 @@
...
+import com.palantir.gradle.testing.execution.TaskOutcome;
+import com.palantir.gradle.testing.junit.GradlePluginTests;
+import com.palantir.gradle.testing.project.RootProject;
+import com.palantir.gradle.testing.project.SubProject;
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+class InvocationResultUsagesTest {
+    @Test
+    void a_failing_build_throws_an_error_when_it_was_expected_to_succeed(
```
crogoz, line 34: For me Intellij is complaining that for all these tests that `Method ... annotated with '@Test' should not declare parameters 'gradle', 'rootProject' and 'rootProject'`

Screenshot 2025-10-13 at 11 00 41 AM

Should we disable this inspection ("JUnit malformed declaration" for all GradlePluginTests) in an Intellij plugin ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/RootProjectStore.java
```diff
@@ -0,0 +1,76 @@
...
+            return projectDir;
+        });
+    }
+
+    private static void clearDirectory(Path projectDir) {
+        try {
+            FileUtils.deleteDirectory(projectDir.toFile());
+            Files.createDirectories(projectDir);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
crogoz:
```suggestion
            throw new UncheckedIOException("Failed to recreate the test directory", e);
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/TerseParameterResolver.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.jupiter.api.extension.ParameterContext;
+import org.junit.jupiter.api.extension.ParameterResolutionException;
+import org.junit.jupiter.api.extension.ParameterResolver;
+
+interface TerseParameterResolver extends ParameterResolver {
+    Optional<Object> parameter(ParameterContext parameterContext, ExtensionContext extensionContext)
+            throws ParameterResolutionException;
+
+    @Override
+    default boolean supportsParameter(ParameterContext parameterContext, ExtensionContext extensionContext)
```
crogoz, line 30: Do we plan on supporting @ParameterizedTests as well ? From what I've seen we can pass an extra parameter to a test but it needs to be the first one but then probably we might want the project directory name to be customized based on the extra parameters ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
```diff
@@ -0,0 +1,69 @@
...
+
+    default GradleSourceSet mainSourceSet() {
+        return sourceSet("main");
+    }
+
+    default GradleSourceSet testSourceSet() {
+        return sourceSet("test");
+    }
+
+    default GradleSourceSet sourceSet(String sourceSetName) {
```
kelvinou01, line 67: I think it'd be helpful to standardize method naming formats here:

Methods which create things  start with verbs:
```java
project.addSubproject("mySubproject");
project.createSourceSet("myCustomSourceSet");
```

Methods which get things have no prefix:
```java
project.buildGradle();
project.mainSourceSet();
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
kelvinou01, line 30: This is nice!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/RootProjectStore.java
```diff
@@ -0,0 +1,76 @@
...
+                Files.writeString(
+                        projectDir.resolve("settings.gradle"),
+                        "rootProject.name = 'root'\n",
+                        StandardOpenOption.CREATE);
+            } catch (IOException e) {
+                throw new UncheckedIOException("Could not create settings.gradle", e);
+            }
+
+            return projectDir;
+        });
```
felixdesouza: you can just add `Path.class` to the end here:

    <K, V> V getOrComputeIfAbsent(K key, Function<K, V> defaultCreator, Class<V> requiredType)

negates the need to cast

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/RootProjectStore.java
```diff
@@ -0,0 +1,76 @@
...
+            clearDirectory(projectDir);
+
+            try {
+                Files.writeString(
+                        projectDir.resolve("settings.gradle"),
+                        "rootProject.name = 'root'\n",
+                        StandardOpenOption.CREATE);
+            } catch (IOException e) {
+                throw new UncheckedIOException("Could not create settings.gradle", e);
+            }
```
felixdesouza: This feels like the wrong place for it to happen. I would have thought the side effects would happen in a `BeforeEachCallback#beforeEach` method perhaps where the parameter resolver just included the location. But I can see that it's used in multiple parameter resolvers so idk. I was just thrown off by this bit and took a minute to reason through it. Key bit here is the `getOrComputeIfAbsent` I guess

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/GradleVersionsFromJunitParameterTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.platform.engine.TestExecutionResult.Status;
+import org.junit.platform.engine.discovery.DiscoverySelectors;
+import org.junit.platform.testkit.engine.EngineExecutionResults;
+import org.junit.platform.testkit.engine.EngineTestKit;
+import org.junit.platform.testkit.engine.Event;
+
+final class GradleVersionsFromJunitParameterTest {
+    @Test
+    void runs_tests_with_gradle_versions_from_junit_parameter() {
+        EngineExecutionResults executionResults = EngineTestKit.engine("junit-jupiter")
```
felixdesouza, line 35: idk why but I really enjoy writing junit engine tests! I had to do a fair bit when debugging our intenal junit5 globalsingleton extension

That said, they provide a number of helpers that are quite neat. We should use the `EventConditions` class and pass these conditions on the events to the assertions. I think it's a bit more declarative, and shows you what you need to know without having to look inside methods like a matryoshka doll.

Here's something I was playing with on Friday:

```java
void test() {
    // preamble
    executionResults
                .testEvents()
                .debug()
                .assertEventsMatchLooselyInOrder(
                        event(
                                test("testName"),
                                runWithGradleVersion("7.6.5"),
                                finishedWithFailure(messageContaining(gradleVersionDisplayName("7.6.5")))),
                        event(
                                test("testName"),
                                runWithGradleVersion("8.14.3"),
                                finishedWithFailure(messageContaining(gradleVersionDisplayName("8.14.3")))));
}

// does not have to be this way, we can just have a condition on a container event that has the display name. This already works without new code i.e. `container(displayName("Gradle 7.6.5"))`
private static Condition<Event> runWithGradleVersion(String gradleVersion) {
    return new Condition<>(
            where(
                    event -> event.getTestDescriptor()
                            .getParent()
                            .map(TestDescriptor::getDisplayName)
                            .orElse(""),
                    Predicate.isEqual(gradleVersionDisplayName(gradleVersion))),
            "parent descriptor run with gradle version '%s'",
            gradleVersion);
}

private static Condition<Throwable> messageContaining(String substring) {
    return new Condition<>(
            where(Throwable::getMessage, message -> message.contains(substring)),
            "message contains '%s'",
            substring);
}

private static String gradleVersionDisplayName(String gradleVersion) {
    return "GradleVersion: " + gradleVersion;
}

```

In fact it doesn't all have to be relevant to a test event. I had previously had it such that the `container` i.e. an instantiation of the test template with the gradle version was what was matched. So you end up with a timeline of events i.e.:

* in gradle 7.6.5 nested container
* in test that finished with failure "gradle 7.6.5"
* in gradle 8.14.3 nested container
* in test that finished with failure "gradle 8.14.3"

this is without the custom condition `runWithGradle`. Either way, making conditions that make our test a bit more readable isn't the worst thing in the world

I think as time goes on and we start to add more functionality, the existing assertions you have will be modified and lose meaning

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
CRogers: Let's just leave it as `root`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradlePluginTests.java
CRogers: Yes! This was a big reason I wanted to use this approach:

Screenshot 2025-10-13 at 16 09 23
Screenshot 2025-10-13 at 16 09 20

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleVersioningClassTemplate.java
CRogers: Yeah, I started writing this in March when it was in beta, nice it's now released

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java
CRogers: Yeah, it's kinda annoying. It's not malformed so their check is wrong. I'm going to file a ticket on them. But you can suppress this check - in fact we could get the plugin to suppress it I think if we really needed to/they don't fix the bug.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradlePluginTests.java
CRogers: That said, as @crogoz pointed out, I need to make one test dir per test instance properly

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/GradleVersionsFromJunitParameterTest.java
```diff
@@ -0,0 +1,79 @@
...
+            assertThatTestFailureExceptionMessageContains(testExecutionResult, "GradleVersion: " + gradleVersion);
+        });
+
+        assertThat(Path.of(
+                        "build/gradle-testing",
+                        MultipleGradleVersionsFixtureTest.class.getSimpleName(),
+                        "testName",
+                        gradleVersion,
+                        "build.gradle"))
+                .exists();
```
felixdesouza, line 63: another thing you can do is use `TestReporter`. I'm not sure of any thing consuming the reports that `TestReporter` in gradle land, but it makes for nice tests:

```java
@Test
void test(RootProject rootProject, TestReporter testReporter) {
  // preamble
  testReporter.publishEntry("root-settings-gradle", "path/to/it");

  testReporter.publishFile(rootProject.settingsGradle().path(), MediaType.TEXT_PLAIN);
}
```

You could write a file in a known location *within* the build i.e. `./gradlew writeGradleVersionToFile`. It'd be within `build/gradleVersion`. You could then publish that file directly here and assert on its contents.

How would you actually consume this:

```java
assertEventsMatchLooselyInOrder(

  event(reportEntry(Map.of("root-settings-gradle", "path/to/it")),
  event(fileEntry(entry -> Files.readString(entry.getPath).contains("rootProject.name = ")
```

It's a nicer way to pass things between the harness and the method under test rather than through exceptions. Granted exception messages can also be handled well as well

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
felixdesouza: don't really have an opinion, but curious why the `IntegrationSpec` naming it test directory is so bad?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/GradleVersionsFromJunitParameterTest.java
CRogers: > idk why but I really enjoy writing junit engine tests!

I'm glad someone does lol

> That said, they provide a number of helpers that are quite neat

yes, that's much nicer, thanks for the tips

CRogers: The only issue is the use of condition rather than individual assertThats makes it super hard to see what was the error

CRogers: Yeah, having this massive block makes it very hard to work out which of the many conditions are actually the problem:

```
	Event [type = FINISHED, testDescriptor = TestMethodTestDescriptor: [engine:junit-jupiter]/[class-template:com.palantir.example.MultipleGradleVersionsFixtureTest]/[class-template-invocation:#1]/[method:testName(com.palantir.gradle.testing.execution.GradleInvoker, com.palantir.gradle.testing.project.RootProject)], timestamp = 2025-10-13T16:56:27.031125Z, payload = TestExecutionResult [status = FAILED, throwable = java.lang.RuntimeException: 


-- failure 1 --Condition did not match any event: all of:[
   all of:[
      is a test,
      descriptor with uniqueId substring 'testName'
   ],
   parent descriptor run with gradle version '7.6.5',
   all of:[
      type is FINISHED,
      event with result where all of:[
   status is FAILED,
   all of:[
      throwable matches message contains 'GradleVersion: 7.6.5'
   ]
]
   ]
]
```

as opposed to just being told exactly which of these many conditions are correct. Reminds me of using hamcrest vs the much better error messages from assertj

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradlePluginTests.java
CRogers: which I have now done

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/TerseParameterResolver.java
CRogers: Good catch, you were right, I had to make new dirs for them too. I've done a more general naming for the test dirs such that it should handle any junit context path. It does work but as you say needs to be the first param, which isn't too bad imo.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
CRogers: Agreed, but I think the `add`/`create` are superfluous so I'll remove them.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/RootProjectStore.java
CRogers: Yeah, imo it would be far worse for this to be created "invalid"

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/RootProjectStore.java
CRogers: done

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java
CRogers: ...if we produced assertions class for this (in a later PR?)

CRogers:
This is ready for review again

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -0,0 +1,73 @@
...
+        assertThat(gradle.withArgs().buildsSuccessfully().output()).contains("hello from :subProject");
+    }
+
+    @Test
+    void two_layer_deep_sub_project(GradleInvoker gradle, SubProject subProject) {
+        SubProject subSubProject = subProject.subproject("subSubProject");
+
+        subSubProject.buildGradle().append("""
+            println "hello from ${path}"
+            """);
```
crogoz: nit nit nit - I am a bit biased towards to the "addSubproject" naming - especially in these cases where other variables might be named similarly. Sometimes it might not be too clear if we create a new subProject/we are fetching it. Also, maybe an AI might understand an "addSubproject" better - in case we plan to use it in the future.

I also like from [nebula](https://github.com/nebula-plugins/nebula-test/blob/d8767635150adadef66994b71ac39bfc589a7bb6/src/main/groovy/nebula/test/IntegrationTestKitBase.groovy#L73) how I can configure the build.gradle file directory when I create a subproject, most of the times I want to also configure the build.gradle contents for a subproject. Bu this can also be added in a future PR.

gradle-plugin-testing/build.gradle
```diff
@@ -32,12 +33,17 @@ gradlePlugin {
     }
 }
 
-test {
+tasks.named('test', Test) {
     systemProperty 'projectVersion', project.version
 
+    // We need to set the versions of gradle to test aganist as we don't have the gradle plugin installed
```
esword: nit: typo

gradle-plugin-testing-junit/build.gradle
```diff
@@ -0,0 +1,36 @@
...
+    api 'org.junit.jupiter:junit-jupiter-api'
+    api 'org.assertj:assertj-core'
+
+    api 'org.jetbrains:annotations'
+    // Just so people can get Gradle autocomplete in text blocks in intellij injected with the
+    // Gradle language fragments
+    compileOnlyApi gradleApi()
+
+    // VERY INTENTIONALLY we are not exposing test kit as api. We want people to use our own types and not the
+    // test kit types directly. This also means we do expose test kit types in our API and have wrappers for all
```
esword: nit: do or do not expose test kit types in our API?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    public GradleFile append(@Language("Gradle") String text) {
+        return ProjectFile.super.append(text);
+    }
+
+    @Override
+    public GradleFile appendLine(@Language("Gradle") String line) {
+        return ProjectFile.super.appendLine(line);
+    }
+}
```
esword, line 47: For future - could add helper methods for things like dependencies like are in TestContentHelpers

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.testing.files.gradle;
+
+import com.palantir.gradle.testing.files.ProjectFile;
+import java.nio.file.Path;
+import org.intellij.lang.annotations.Language;
+
+public record GradleFile(Path path) implements ProjectFile<GradleFile> {
```
FinlayRJW: can we add an override for edit to get `@Language` on it?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java
```diff
@@ -0,0 +1,183 @@
...
+            println 'hello from build'
+            """);
+
+        InvocationResult result = gradle.withArgs().buildsSuccessfully();
+
+        assertThat(result.output()).contains("hello from build");
+    }
+
+    @Test
+    void non_exsistent_tasks_are_returned_as_optional_empty(GradleInvoker gradle) {
```
esword: nit: typo

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
CRogers: yeah, reading this now `subProject.subproject(` is maybe not the clearest. My thought was - just like the source sets - should you have to care whether you're adding the subproject for the first time or whether it's already been created? We should just do the right thing and let you get hold of a subproject.

re: adding the build gradle code right away - you can do this:

```java
rootProject.subproject("sub").buildGradle().append("""
```

which I think might be good enough?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
CRogers: I want to move away from helpers like that, ideally everyone just uses stuff injected by the JUnit extension.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
CRogers: I think for your use case adding a `prepend` method might just help?

It's a little harder to do, at least you can just have a type extending `ProjectFile` that has `edit(GradleFileEditor)` overriding of `edit(FileEditor)`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
esword: "ideally everyone just uses stuff injected by the JUnit extension" - what's the definition of JUnit extension in this situation?  Meaning what do you define as being injected by that vs added to file contents with methods like above?

I brought up the idea because of the custom addValue in the PropertiesFile.

CRogers: Ah, completely misread - I thought you meant make these `*File` types available from the jar that has the `TestContentHelpers`. Yep, being able to add dependencies/plugins here (especially plugins to test against that we keep upgraded) would be great!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
CRogers: I tried this and while it's possible, I don't think it's worth it. Since the `@Language` is being applied on a _return_ value, there are only limited situations where this is injected by intellij, notably string constants or concatenation of string constants, and all of these have better methods to use:

* Just returning a string constant:
    ```java
    file.edit(_ignored -> "new text")
    ```
    This is just `overwirte("new text")` but worse.
* Appending a string constant:
    ```java
    file.edit(text -> text + "at end")
    ```
    This is just `append("at end")` but worse.
* Prepending a string constant:
    ```java
    file.edit(text -> "at front" + text)
    ```
    This is common enough we could just add a method `prepend("at front")`.
* More complex concatenation:
    ```java
    file.edit(text -> "at front" + text + "at end")
    ```
    This can just be a combo of `file.prepend("at front").append("at end")`.

I'll at least make a `FileEditor` functional interface so this is a bit less of a breaking change if we want to introduce it.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/java/JavaSrcDir.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.nio.file.Path;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.intellij.lang.annotations.Language;
+
+public record JavaSrcDir(Path srcDirPath) {
+    private static final Pattern PACKAGE_PATTERN = Pattern.compile("package\\s+([^;]+);");
+    private static final Pattern CLASS_PATTERN = Pattern.compile("(?:class|interface|record|enum)\\s+(\\w+)");
```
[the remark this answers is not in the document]
CRogers: `interface` matches `@interface` which is why the test works - I guess we should still add it for clarity.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
FinlayRJW: yeah that makes sense I think a `prepend` would be useful and cover a lot of cases where you might use `edit`

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
CRogers: I think I'd like to keep it just `subproject` so it's ok to request it multiple times. I've made sure it will do the right thing when you call `subproject` multiple times.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.common.base.Splitter;
+import com.google.errorprone.annotations.RestrictedApi;
+import com.palantir.gradle.testing.RestrictedCreation;
+import java.nio.file.Path;
+import java.util.stream.Collectors;
+import org.intellij.lang.annotations.Language;
+
+public record SettingsGradleFile(Path path) implements GradleFile {
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public SettingsGradleFile {}
```
CRogers, line 29: RestrictedApi allows us to use records (which require a public constructor) but then ban people from actually using it (well, people using our errorprone infra).

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private String projectNameFromParameter(ParameterContext parameterContext) {
+        String paramName = parameterContext.getParameter().getName();
+
+        if (paramName.endsWith("Project")) {
+            return paramName.substring(0, paramName.lastIndexOf("Project"));
+        }
+
+        return paramName;
+    }
```
CRogers, line 54: I've made a change where if you inject a project with the parameter name ending in `Project`, it will strip this off. It makes it a bit easier to read imo, and allows naming the root project easier. Eg:

```java
void test(SubProject serviceProject) {
```

...would make a project called `service`.

CRogers: This means you can rename the root project easily - but I bet 98% of the time people will just use `RootProject rootProject`, making it called `root`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    default T prepend(String text) {
+        edit(existingText -> text + existingText);
+        return (T) this;
+    }
+
+    default T prependLine(String line) {
+        return prepend(line + "\n");
+    }
```
CRogers, line 54: Added `prepend`/`prependLine`

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
CRogers: It meant if you called you test `def 'the distTar task makes a dist tar of the compile source code()', Gradle makes the root project name `the-distTar-task-makes-a-dist-tar-of-the-compile-source-code' (as there was no explicit name set, it uses the parent dir). So now you end up with artifacts called `the-distTar-task-makes-a-dist-tar-of-the-compile-source-code.tgz`. People would just use these long strings in the tests rather than name the root project. Now if you change the test name, everything breaks.

CRogers: I also changed my mind here and people can name `rootProject` from the parameter. But 98% of the people will use `rootProject`, making it `root`.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java
felixdesouza: sorry yes, this was in relation to my other comment re assertions classes (not in this PR)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
```diff
@@ -0,0 +1,48 @@
...
+package com.palantir.gradle.testing.files;
+
+import com.palantir.gradle.testing.files.arbitrary.ArbitraryFile;
+import com.palantir.gradle.testing.files.gradle.GradleFile;
+import com.palantir.gradle.testing.files.gradle.RegularGradleFile;
+import com.palantir.gradle.testing.files.properties.PropertiesFile;
+import com.palantir.gradle.testing.files.yaml.YamlFile;
+import java.nio.file.Path;
+
+public interface FileFactory {
```
felixdesouza, line 26: not blocking, but considering that you're blocking calls, would you make this a sealed interface? not fussed either, mostly curious from a conceptual pov

CRogers: tbh I had just completely forgotten that was tool available to us. Nice idea though, I will try it out!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
felixdesouza: don't you always want to add a new line? I would have thought `prepend` by itself is a little strange in that you're effectively "guessing" what's at the beginning and tacking that on and hoping it's okay? To a lesser extent, the same is probably true for `append` and `appendLine` ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
CRogers: Unfortunately:

> Record 'com.palantir.gradle.testing.files.arbitrary.ArbitrarySrcDir' from another package not allowed to extend sealed interface 'com.palantir.gradle.testing.files.FileFactory' in unnamed module

...which is a bit annoying. Perhaps more fuel for the "everything in the same package argument". Seems a bit of a silly constraint imo.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
```diff
@@ -0,0 +1,55 @@
...
+            return Optional.of(rootProjectFor(parameterContext, extensionContext)
+                    .subproject(projectNameFromParameter(parameterContext)));
+        }
+
+        return Optional.empty();
+    }
+
+    private RootProject rootProjectFor(ParameterContext parameterContext, ExtensionContext extensionContext) {
+        RootProject rootProject = RootProjectStore.rootProject(extensionContext);
+        rootProject.settingsGradle().rootProjectName(projectNameFromParameter(parameterContext));
```
felixdesouza, line 42: I'm a bit confused why we have to set the `rootProjectName` here and also in `RootProjectStore`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
CRogers: Text blocks have a new line at the end (unless you end their content with a backslash eg `foo\`). So you'd have to check for the newline + always add it or just be ok adding two newlines. Both of these are probably fine for Gradle files - not necessarily true for all file types.

These `*Line` methods are really so you can use non-text-block strings without having to add the newline character, eg `file.appendLine("blah")` instead of `file.append("blah\n")`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
CRogers: I guess I was concerned about making a `RootProject` object with no default name set - but since the the only way to get a `RootProject` is via the parameter resolver, we can get away with setting it once.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
felixdesouza: even with the `permits` clause? I agree it's a very weird constraint if so

CRogers: Yes, that was with `permits ArbitrarySrcDir`

CRogers: The only reason I can think is they were being very conservative when implementing it with the possibility of expanding later.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
```diff
@@ -0,0 +1,70 @@
...
+import java.nio.file.Path;
+
+public interface GradleProject extends FileFactory {
+    @Override
+    Path path();
+
+    RootProject rootProject();
+
+    default SubProject subproject(String name) {
+        Path subprojectDir = path().resolve(name);
```
felixdesouza, line 39: something I ran into in GCV, we should enforce that the name provided here is not a project path i.e. `foo:a:b`. Either that or split and create this subproject hierarchy, but I think the former is an easier place to start.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
CRogers: Apparently if you use modules you can have multiple packages. But that's a whole unexplored area.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ParameterizedTestUsagesTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.jupiter.params.provider.ValueSource;
+
+@GradlePluginTests
+final class ParameterizedTestUsagesTest {
+    @Nested
+    class NestedClass {
+        @ParameterizedTest(name = "{index}: {0}")
+        @ValueSource(strings = {"foo", "bar", "baz"})
+        void a_test_with_params_has_one_directory_per_param(String param, RootProject rootProject) {
+            // TODO(callumr): Actually test this properly
```
felixdesouza: is this for this PR or something else?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
CRogers: I banned colons and `/` from the project path. Other invalid characters I think Gradle can check well enough.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
CRogers: Turns out you do need a `settings.gradle` otherwise you get this error: `Build cache controller already set`. This was happening in the few tests in this repo where `RootProject` was not injected. So I am creating it again. I'm not sure consumers will ever do this though - it's just running a blank build with no customisation at all - why would anyone do that.

felixdesouza: oh that's weird, but fair enough

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
felixdesouza: yeah that's fine, it's more for https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#project_directories_must_exist_and_be_writeable than anything else i.e. as long as there's a 1-1 mapping from name to directory, not fussed if the name is invalid

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ParameterizedTestUsagesTest.java
CRogers: Yeah, I though I had to do a junit-test-kit test for this, but I don't

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
felixdesouza: that will hopefully remain unexplored 😂

felixdesouza:
👍  🚀

PR #221, opened 2025-10-20
Add errorprone infra; also add errorprone to ban manual temporary file/directory creation

CRogers:
## Before this PR
We have no way to make error prones for this project. I want us to be able to make error-prone as and when we spot them quickly, so having the infra already setup would be a great aid.

Additionally, in [this PR](https://github.com/palantir/gradle-idea-language-injector/pull/12#discussion_r2444888266) there was an attempted use of temporary directories. I think temporary file/directory use like this is an antipattern - each gradle plugin test has it's own directory to do stuff in that persists after the gradle run has finished. People should just use this directly rather than some temporary directory that is deleted after the test runs, meaning it's very hard to see what was in it.

## After this PR
==COMMIT_MSG==
Add errorprone to ban manual temporary file/directory creation in favour of just making a file/directory in the gradle test root directory.
==COMMIT_MSG==


felixdesouza:
👍  neat!

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestTemporaryFile.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        return describeMatch(tree);
+    }
+
+    private static boolean withinGradlePluginTestsClass(VisitorState state) {
+        return StreamSupport.stream(state.getPath().spliterator(), false)
+                .anyMatch(parentTree -> ASTHelpers.hasAnnotation(
+                        parentTree, "com.palantir.gradle.testing.junit.GradlePluginTests", state));
+    }
```
kelvinou01: This might be more concise, and it only matches on the first enclosing class.
 
`Matchers.enclosingClass(Matchers.hasAnnotation("com.example.MyAnnotation"))`

gradle-plugin-testing-error-prone/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+            'jdk.compiler/com.sun.tools.javac.comp',
+            'jdk.compiler/com.sun.tools.javac.main',
+            'jdk.compiler/com.sun.tools.javac.model',
+            'jdk.compiler/com.sun.tools.javac.parser',
+            'jdk.compiler/com.sun.tools.javac.processing',
+            'jdk.compiler/com.sun.tools.javac.tree',
+            'jdk.compiler/com.sun.tools.javac.util'
+    ]
+    opens = ['jdk.compiler/com.sun.tools.javac.comp']
+}
```
kelvinou01, line 32: ok now i see the need for https://github.com/palantir/suppressible-error-prone/issues/178 😬

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestTemporaryFile.java
```diff
@@ -0,0 +1,88 @@
...
+import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.VariableTree;
+import java.util.stream.StreamSupport;
+
+@AutoService(BugChecker.class)
+@BugPattern(severity = SeverityLevel.ERROR, summary = """
+    Avoid using temporary directories or files in Gradle tests, as it can be hard to view the \
+    contents of it for debugging. Instead, just make a directory or file using `RootProject` or \
+    `SubProject`. These will remain around after the test completes in the `build/gradle-plugin-testing` \
+    directory, aiding debugging.
```
kelvinou01:
```suggestion
    Avoid using temporary directories or files in Gradle tests, as it can be hard to view the \
    contents for debugging. Instead, make a directory or file using `RootProject` or \
    `SubProject`, which will remain in `build/gradle-plugin-testing` after the test completes.
```

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestTemporaryFile.java
CRogers: Nice, I didn't know about that. I don't have a test for it, but that's not going to be enough if someone has a nested class like this:

```java
@GradlePluginTests
class Outer {
    @Nested
    class Nested {
        @Test
        void test() {
```

kelvinou01: Ahh. How about

```java
Matchers.enclosingNode(
    Matchers.allOf(
          Matchers.isInstance(ClassTree.class),
          Matchers.hasAnnotation("com.palantir.gradle.testing.junit.GradlePluginTests")
      )
  );
```

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestTemporaryFile.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (!MANUAL_TEMPORARY_METHOD_MATCHER.matches(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        boolean withinGradlePluginTestsClass = withinGradlePluginTestsClass(state);
+
+        if (!withinGradlePluginTestsClass) {
```
FinlayRJW: nit: inline this as `if (!withinGradlePluginTestsClass(state)`

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestTemporaryFile.java
CRogers: yup, that does it

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestTemporaryFile.java
```diff
@@ -0,0 +1,84 @@
...
+                    "com.google.common.io.Files",
+                    "org.apache.commons.io.FileUtils")
+            .namedAnyOf(
+                    "createTempDir",
+                    "createTempDirectory",
+                    "createTempFile",
+                    "getTempDirectory",
+                    "getTempDirectoryPath");
+
+    private static final Matcher<Tree> WITHIN_GRADLE_PLUGIN_TESTS_CLASS = Matchers.enclosingNode(Matchers.allOf(
```
FinlayRJW, line 55: I'm guessing that all future errorprone will need to have this check - could we extract some base `GradlePluginTestBugChecker` which handles this so we don't end up copying everywhere?

FinlayRJW: Although we could always do this in the next PR to add more errorprones

CRogers: Yeah, we should do something like that (probably just make a static method for it) when we add the code in the next PR.

FinlayRJW:
LGTM 👍

PR #233, opened 2025-10-23
Add custom AssertJ assertions

FinlayRJW:
## Before this PR

  Test assertions for Gradle invocations were verbose and not fluent - developers had to manually extract task results and write assertion logic using nested lambdas.

  ## After this PR

  ### Fluent Assertion API on InvocationResult

  Added fluent assertion methods directly to `InvocationResult` with comprehensive test coverage.

  **InvocationResultAssert** - Entry point for fluent assertions:
  - `assertThat()` - Returns fluent assertion chain for invocation results
    - `task(String taskPath)` - Returns `TaskResultOptionalAssert` for task assertions
    - `output()` - Returns AssertJ `StringAssert` for build output assertions

  **TaskResultOptionalAssert** - Fluent assertions for task results:
  - `outcome()` - Returns `TaskOutcomeAssert` for outcome assertions
  - `wasSuccess()` - Convenience method to assert task succeeded
  - `wasFail()` - Convenience method to assert task failed
  - `wasUpToDate()` - Convenience method to assert task was up-to-date
  - `isEmpty()` - Assert task was not executed

  **TaskOutcomeAssert** - Fluent assertions for task outcomes:
  - `wasSuccess()` - Assert outcome is SUCCESS
  - `wasFail()` - Assert outcome is FAILED
  - `wasUpToDate()` - Assert outcome is UP_TO_DATE
  - `isEqualTo()` - Assert exact outcome match (supports all TaskOutcome values)

  **Example usage:**
  ```java
  // Before: Verbose nested lambdas
  assertThat(result.task(":foo")).hasValueSatisfying(taskResult -> {
      assertThat(taskResult.outcome()).isEqualTo(TaskOutcome.SUCCESS);
  });

  // After: Fluent chain
  result.assertThat().task(":foo").outcome().isEqualTo(TaskOutcome.SUCCESS);

  // Or use convenience methods
  result.assertThat().task(":foo").wasSuccess();

  // Output assertions
  result.assertThat().output().contains("BUILD SUCCESSFUL");

  // Check task was not executed
  result.assertThat().task(":notRun").isEmpty();
```

  All existing tests updated to use new fluent API. Comprehensive test coverage added in GradleAssertionsUsageTest.

  ==COMMIT_MSG==
  Add fluent assertion API to InvocationResult

  - Add assertThat() method to InvocationResult with task() and output() chains
  - Add TaskResultOptionalAssert with outcome() and convenience methods (wasSuccess, wasFail, wasUpToDate)
  - Add TaskOutcomeAssert with outcome-specific assertions
  - Update all existing tests to use fluent API
  - Add comprehensive test coverage in GradleAssertionsUsageTest
==COMMIT_MSG==

  Possible downsides?

  None expected. All changes are additive and backward compatible.  ## Before this PR

### CODE REVIEW

File: gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResultAssert.java [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+package com.palantir.gradle.testing.execution;
+
+import java.util.Optional;
+import org.assertj.core.api.OptionalAssert;
+
+final class InvocationResultAssert {
+
+    private InvocationResultAssert() {}
+
+    public static final class TaskResultOptionalAssert extends OptionalAssert<TaskResult> {
```
CRogers: Feels kinda annoying to have this nested inside another class - was the idea that there would be other assertions in here at some point?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResult.java
```diff
@@ -33,4 +35,30 @@ public String output() {
...
+     * Usage:
+     * <pre>
+     * invocationResult.assertTask(":myTask")
+     *     .hasOutcome()
+     *     .isNotIn(TaskOutcome.UP_TO_DATE, TaskOutcome.FROM_CACHE);
+     * </pre>
+     */
+    public InvocationResultAssert.TaskResultOptionalAssert assertTask(String taskPath) {
+        return new InvocationResultAssert.TaskResultOptionalAssert(task(taskPath));
+    }
```
CRogers: I wonder if instead of `assertTask` and `assertOutput` we had an `assertThat()` method that people could do `.assertThat().task(":path")...` and `.assertThat().output()`. Would stop methods proliferating on this class, and I think it's easier for people to pattern match on `assertThat`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResultAssert.java
FinlayRJW: yeah I used to have a `hasOutput` in here but was moved to `InvocationResult#assertOutput`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResultAssert.java
```diff
@@ -0,0 +1,60 @@
...
+            super(optional);
+        }
+
+        @Override
+        public TaskResultOptionalAssert as(String description, Object... args) {
+            return (TaskResultOptionalAssert) super.as(description, args);
+        }
+
+        public TaskOutcomeAssert hasOutcome() {
+            isPresent();
```
CRogers: Not convinced this going to give the best error messages. Generally using `hasValueSatisfying` gives better errors. I wonder if we can use that codepath. It might make the api a bit weird though, might need some finagling.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/execution/GradleAssertionsTest.java
```diff
@@ -0,0 +1,136 @@
...
+package com.palantir.gradle.testing.execution;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import com.palantir.gradle.testing.junit.GradlePluginTests;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+class GradleAssertionsTest {
```
CRogers: I'd call this a `UsagesTest` in line with the other tests that actually use the testing framework directly.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResult.java
CRogers: I think this may also fix the "nested class" comment [here](https://github.com/palantir/gradle-plugin-testing/pull/233#discussion_r2455621172)

FinlayRJW: yeah that is much nicer

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResultAssert.java
FinlayRJW: I'm not sure about this - both `isPresent` and `hasValueSatisfying` call into `assertValueIsPresent` which both give `Expecting Optional to contain a value but it was empty` - if we want a custom error message we can do something more like 
```java
    public TaskOutcomeAssert hasOutcome() {
        isNotNull();
        if (!actual.isPresent()) {
            failWithMessage("Expected TaskResult to contain a task outcome but was empty.");
        }
        return new TaskOutcomeAssert(actual.get().outcome());
    }
```
if you think that would be nicer?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/InvocationResultAssert.java
```diff
@@ -0,0 +1,56 @@
...
+    public InvocationResultAssert(InvocationResult invocationResult) {
+        this.invocationResult = invocationResult;
+    }
+
+    /**
+     * Returns an assertion object for fluent task assertion chaining.
+     * <p>
+     * Usage:
+     * <pre>
+     * invocationResult.assertThat().task(":myTask")
```
crogoz: can we move the existing tests (eg. [here](https://github.com/palantir/gradle-plugin-testing/blob/8e30c544432fa28ef521a147ecf9664bf5f67ec2/gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/InvocationResultUsagesTest.java#L88-L89)) to this new API ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskOutcomeAssert.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.stream.Collectors;
+import org.assertj.core.api.AbstractObjectAssert;
+
+public final class TaskOutcomeAssert extends AbstractObjectAssert<TaskOutcomeAssert, TaskOutcome> {
+
+    TaskOutcomeAssert(TaskOutcome taskOutcome) {
+        super(taskOutcome, TaskOutcomeAssert.class);
+    }
+
+    public TaskOutcomeAssert isIn(TaskOutcome... expectedOutcomes) {
```
crogoz: doesn't 1 task have only 1 outcome ? when would we want to check if it matches any of a list of Outcomes ? eg. I could have `.hasOutcome().isIn(TaskOutcome.UP_TO_DATE, TaskOutcome.FAILED);` which would give true

FinlayRJW: true - I created this based off my use case here (https://github.com/palantir/gradle-idea-language-injector/pull/12/files#diff-8f90a8275f55733a1f82d4a1f3a44fd279a04b4ad794b5d17c0143f2ee361697R174) but its probably better for me to just say `hasOutcome(x)` - honestly the same could be said for isNotIn - i.e. we always force you to say exactly what the task outcome should be rather than what its not

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultOptionalAssert.java
```diff
@@ -0,0 +1,52 @@
...
+    TaskResultOptionalAssert(Optional<TaskResult> optional) {
+        super(optional);
+    }
+
+    @Override
+    public TaskResultOptionalAssert as(String description, Object... args) {
+        return (TaskResultOptionalAssert) super.as(description, args);
+    }
+
+    public TaskResultOptionalAssert hasOutcome(TaskOutcome expected) {
```
CRogers: Tasks always have an outcome - I think it would be more readable if it was just `.outcome()`. `assertThat().task("foo").hasOutcome().isEqualTo(TaskOutcome.SUCCESS)` perhaps sounds a bit weird compared to `assertThat().task("foo").outcome().isEqualTo(TaskOutcome.SUCCESS)` even that doesn't really roll of the tongue to be fair.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResultAssert.java
CRogers: Yeah, I think we should have a custom error message. Tbh haven't written any custom assertj assertion classes so this is new to me.

Do we need the `isNotNull()` call? It appears to check the `Optional<TaskResult>` is not null, which should never be the case. I think we'd just need the `if (!actual.isPresent`

> Expected TaskResult to contain a task outcome but was empty.

Task results always contain an outcome. I think this is really checking that there was a task with that name.  So something like "Expected to find a task result for task ':path' but there was none"?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResultAssert.java
CRogers: Can we just make this class top level now?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultOptionalAssert.java
FinlayRJW: so the current api is more:

```java
secondRun.assertThat()
    .task(":myTask")
    .hasOutcome(TaskOutcome.UP_TO_DATE);
```

Where you pass into hasOutcome - we could add or just have an outcome method that returns the optional then you can do your own checks on it? i.e. 

```java
secondRun.assertThat()
    .task(":myTask")
    .outcome().isEqualTo(TaskOutcome.SUCCESS);
```

the second is more similar to nebula so maybe we want to go that route?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResultAssert.java
FinlayRJW: yep I don't this we need the `isNotNull` have added our own check with better error message

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultOptionalAssert.java
FinlayRJW: so I've switched to the second style, it's more inline with what nebula did and I think generally clearer

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResult.java
```diff
@@ -33,4 +34,20 @@ public String output() {
...
     }
+
+    /**
+     * Returns an assertion object for fluent assertion chaining.
+     * <p>
+     * Usage:
+     * <pre>
+     * invocationResult.assertThat().task(":myTask")
+     *     .hasOutcome()
+     *     .isNotIn(TaskOutcome.UP_TO_DATE, TaskOutcome.FROM_CACHE);
```
crogoz:
```suggestion
     *     .outcome()
     *     .isEqualTo(TaskOutcome.UP_TO_DATE);
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/InvocationResultAssert.java
```diff
@@ -0,0 +1,56 @@
...
+     *     .hasOutcome()
+     *     .isNotIn(TaskOutcome.UP_TO_DATE, TaskOutcome.FROM_CACHE);
```
crogoz:
```suggestion
     *     .outcome()
     *     .isEqualTo(TaskOutcome.UP_TO_DATE);
```

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/GradleAssertionsUsageTest.java
```diff
@@ -0,0 +1,142 @@
...
+                doLast {}
+            }
+            """);
+
+        InvocationResult result = gradle.withArgs("foo").buildsSuccessfully();
+
+        result.assertThat()
+                .task(":foo")
+                .as("Task should have correct path")
+                .path()
```
CRogers: Do we think we'll ever need this `.path()` assertion? You need to give the path to get the task assertion - I'm not sure when you'd need it? Maybe we should just leave it out until people do need it?

FinlayRJW: yeah fair I was matching the api from nebula but I can't see much of a usecase

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskOutcomeAssert.java
```diff
@@ -0,0 +1,51 @@
...
+import com.palantir.gradle.testing.execution.TaskOutcome;
+import org.assertj.core.api.AbstractObjectAssert;
+
+public final class TaskOutcomeAssert extends AbstractObjectAssert<TaskOutcomeAssert, TaskOutcome> {
+
+    TaskOutcomeAssert(TaskOutcome taskOutcome) {
+        super(taskOutcome, TaskOutcomeAssert.class);
+    }
+
+    public TaskOutcomeAssert wasSuccess() {
```
FinlayRJW: FLUP: maybe an errorprone to force the usage of these

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/GradleAssertionsUsageTest.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    @Test
+    void can_check_task_is_empty(GradleInvoker gradle) {
+        InvocationResult result = gradle.withArgs().buildsSuccessfully();
+
+        result.assertThat()
+                .task(":nonexistent")
+                .as("Non-existent task should be empty")
+                .isEmpty();
```
felixdesouza: can we have a more readable helper, what does it mean for a task to be "empty" ? There's a verbose way of saying this as 

    assertThat().task(":nonexistent").didNotRun()

However a bit ambiguous as it could be up-to-date, from cache etc. 

I think the thing you care about is:

    assertThat().task(":nonexistent").notOnTaskGraph()

There probably is a nicer way of phrasing that, but `empty` is too ambiguous plus does not read well either.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultOptionalAssert.java
```diff
@@ -0,0 +1,59 @@
...
+        return new TaskOutcomeAssert(requireTaskResult().outcome());
+    }
+
+    public TaskOutcomeAssert wasSuccess() {
+        return outcome().wasSuccess();
+    }
+
+    public TaskOutcomeAssert wasFail() {
+        return outcome().wasFail();
+    }
```
felixdesouza: these do not read well, i.e. you want to say "task was successful", or "task was failure" either that or:

* "task succeeded" or "task failed"

"task was success" probably *could* be left the same, however "task was fail" does not read well at all

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskOutcomeAssert.java
```diff
@@ -0,0 +1,48 @@
...
+        if (!actual.equals(TaskOutcome.SUCCESS)) {
+            failWithMessage("Expected task outcome to be SUCCESS but was %s", actual);
+        }
```
felixdesouza: why not just use an assertion here? rather than rolling with it yourself?

felixdesouza: plus could likely just make a helper that does this assertion with the `TaskOutcome` passed in.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/InvocationResultAssert.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Returns an assertion object for fluent output assertion chaining.
+     * <p>
+     * Usage:
+     * <pre>
+     * invocationResult.assertThat().output().contains("BUILD SUCCESSFUL");
+     * </pre>
+     */
+    public AbstractStringAssert<?> output() {
```
felixdesouza: I wonder whether we future proof this by making it return a `BuildOutputAssert` so we can add our own methods on it. 

We can leave it as is, but it would be worth checking that changing the return type to something that extends `AbstractStringAssert` does not result in compile or runtime breaks.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskOutcomeAssert.java
```diff
@@ -0,0 +1,48 @@
...
+        return this;
+    }
+
+    public TaskOutcomeAssert wasFail() {
```
felixdesouza: I think this should be: `wasFailure` i.e. "assert that task outcome was failure"

Claude seems to suggest that another option is: "assert that task outcome failed"

i.e. `assertThat(task).outcome().failed()` which is also not bad and I would be open to that. It just depends on what we want to do re methods i.e. do we want `was*` attached to every assertion method, or do we just describe it as is.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/GradleAssertionsUsageTest.java
FinlayRJW: ah isEmpty() comes from the us extending `OptionalAssert<TaskResult>` I'll add a `.notOnTaskGraph()` as that is defiantly cleaner

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultOptionalAssert.java
```diff
@@ -0,0 +1,59 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.testing.assertion;
+
+import com.palantir.gradle.testing.execution.TaskResult;
+import java.util.Optional;
+import org.assertj.core.api.OptionalAssert;
+
+public final class TaskResultOptionalAssert extends OptionalAssert<TaskResult> {
```
felixdesouza: let's just drop the `Optional` from `TaskResultOptionalAssert`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskOutcomeAssert.java
FinlayRJW: I got the idea of using the was* prefix from here https://pl.ntr/2y0 - but honestly it is probably overkill and just `failed` / `cached` is better

FinlayRJW: gone with `failed`, `succeeded` and `upToDate`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultOptionalAssert.java
FinlayRJW: gone with failed, succeeded and upToDate

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultOptionalAssert.java
felixdesouza: if you extend `AbstractOptionalAssert<TaskResultOptionalAssert, TaskResult>` instead, you don't have to do any casting

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultAssert.java
```diff
@@ -0,0 +1,59 @@
...
+import com.palantir.gradle.testing.execution.TaskResult;
+import java.util.Optional;
+import org.assertj.core.api.OptionalAssert;
+
+public final class TaskResultOptionalAssert extends OptionalAssert<TaskResult> {
+
+    private final String taskPath;
+
+    TaskResultOptionalAssert(Optional<TaskResult> optional, String taskPath) {
+        super(optional);
```
felixdesouza: this would then be: `super(optional, TaskResultOptionalAssert.class);`

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/GradleAssertionsUsageTest.java
FinlayRJW: added in https://github.com/palantir/gradle-plugin-testing/pull/233/commits/82e0a55f02da4455af3d1bcf91dc4dff1a94adfb

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultAssert.java
FinlayRJW: fixed in https://github.com/palantir/gradle-plugin-testing/pull/233/commits/06a67024d5172c031c02b23d618b275f831e5b07

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/InvocationResultAssert.java
FinlayRJW: added `BuildOutputAssert`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/TaskResultAssert.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private final String taskPath;
+
+    TaskResultAssert(Optional<TaskResult> optional, String taskPath) {
+        super(optional, TaskResultAssert.class);
+        this.taskPath = taskPath;
+    }
+
+    @Override
+    public TaskResultAssert as(String description, Object... args) {
```
felixdesouza: you can also get rid of this override because you're using `AbstractOptionalAssert`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/assertion/InvocationResultAssert.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.testing.assertion;
+
+import com.palantir.gradle.testing.execution.InvocationResult;
+import org.assertj.core.api.AbstractStringAssert;
+import org.assertj.core.api.Assertions;
+
+public final class InvocationResultAssert {
```
felixdesouza: potentially can do this another time, but still would be good to have each thing be extend `AbstractObjectAssert` so then people can still do assertions on the `InvocationResult` for things not on in this assertion class and fluently.

People can do `satisfies`, people can do `extracting` without necessarily having dedicated support here.

felixdesouza (commented):
I kinda feel like we need a `GradlePluginTestAssertions` class that has the static methods:

```java
class GradlePluginTestAssertions {

  static InvocationResultAssert assertThat(InvocationResult invocationResult);
  static TaskOutcomeResult assertThat(TaskOutcome taskOutcome);
  ...
}
```

There's just something that irks me about:

    result.assertThat().outcome().succeeded()

vs

    assertThat(result).outcome().succeeded()

That's not to get rid of the former, but the latter is a lot more canonical imo.

felixdesouza:
There's still lots of missing javadoc btw

FinlayRJW:
> There's still lots of missing javadoc btw

Think I should have added all in now

FinlayRJW:
FLUP - should files assertThat().contains/hasContent have highlighting and formatting overloads?

CRogers:
:+1:

PR #234, opened 2025-10-23
Add custom Directory API

FinlayRJW:
## Before this PR
There was no first-class support for creating and working with directories as test fixtures.

## After this PR
**Breaking Changes:**
- Renamed `FileFactory` → `Directory` 
- Renamed `ArbitrarySrcDir` → `ArbitraryDirectory`
- All implementing classes updated (`BuildDir`, `GradleProject`, etc.)

**New Functionality:**
- Added `Directory.createDirectories()` method to ensure directory existence
- Added `Directory.directory(String path)` method to create nested directory references
- **DirectoryUsagesTest.java** - Tests for directory operations

**Example usage:**
```java
// Create directories and files within them
Directory subDir = rootProject.directory("src/main/resources");
subDir.createDirectories();
subDir.file("config.properties").overwrite("key=value");
subDir.yamlFile("settings.yaml").overwrite("enabled: true");

// Nested directories
Directory parent = rootProject.directory("parent");
Directory child = parent.directory("child");
child.createDirectories();
child.file("test.txt").overwrite("content");
```

## Possible downsides?
**Breaking changes** - Any code using `FileFactory` or `ArbitrarySrcDir` will need to be updated to use the new names (`Directory` and `ArbitraryDirectory`).


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/Directory.java
```diff
@@ -0,0 +1,34 @@
...
+import com.palantir.gradle.testing.RestrictedCreation;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+public record Directory(Path path) implements FileFactory {
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public Directory {
+        try {
+            Files.createDirectories(path);
```
CRogers: Currently in our framework files don't automatically create themselves when you instantiate the object. There's a particularly pernicious trap in nebula-test where people will write assertions like:

```groovy
then:
file('whatever').exists()
directory('whatever').exists()
```

....which are vacuously true as the file/directory is already created. I'd like to avoid introducing this kind of trap again if possible, which likely means not creating at object creation time.

CRogers: What this does to the usability of this - idk. It might be fine to just not create directories until someone either calls an `ensureExists()` method or tries to write to a file (and the file method ensures the directories are created).

FinlayRJW: I originally had a `create` method but felt it was a bit annoying to use but was not aware of the nebula-test trap so will add a ensureExists() back - most of the time users will probably use `.file` anyway so should be fine

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/Directory.java
```diff
@@ -0,0 +1,38 @@
...
+
+public record Directory(Path path) implements FileFactory {
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public Directory {}
+
+    public Directory ensureExists() {
+        try {
+            Files.createDirectories(path);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
```
CRogers: Best to add a string to exceptions like this eg `Failed to ensure directory exists`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
```diff
@@ -30,6 +30,10 @@ default ArbitraryFile file(String path) {
         return new ArbitraryFile(resolvePath(path));
     }
 
+    default Directory directory(String path) {
+        return new Directory(resolvePath(path)).ensureExists();
```
CRogers: This is just as bad as creating the directories in the constructor! To the user it does the same thing!

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/DirectoryTest.java
```diff
@@ -0,0 +1,95 @@
...
+package com.palantir.gradle.testing.files;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.palantir.gradle.testing.junit.GradlePluginTests;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+class DirectoryTest {
```
CRogers: `UsagesTest`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/Directory.java
```diff
@@ -0,0 +1,38 @@
...
+package com.palantir.gradle.testing.files;
+
+import com.google.errorprone.annotations.RestrictedApi;
+import com.palantir.gradle.testing.RestrictedCreation;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public record Directory(Path path) implements FileFactory {
```
CRogers: I'm wondering now whether `FileFactory` and `Directory` should actually be the same thing. A `FileFactory` necessarily needs to be a directory... It might be quite nice to have the `ensureExists` method on the implementations of `FileFactory` but I'm not sure.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/DirectoryUsagesTest.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.palantir.gradle.testing.files.Directory;
+import com.palantir.gradle.testing.junit.GradlePluginTests;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+class DirectoryUsagesTest {
+
```
CRogers, line 28: Definite need a test that it doesn't create a directory when you create the object.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/DirectoryUsagesTest.java
```diff
@@ -0,0 +1,66 @@
...
+    void can_create_files_in_directory(RootProject rootProject) {
+        Directory dir = rootProject.directory("my-dir");
+
+        dir.file("test.txt").overwrite("hello");
+
+        assertThat(dir.file("test.txt").text()).isEqualTo("hello");
+    }
+
+    @Test
+    void directory_implements_file_factory(RootProject rootProject) {
```
CRogers: I don't think this is a necessary test, or at least tests too much of `FileFactory`.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/DirectoryUsagesTest.java
CRogers: Ah, missed the test at the bottom - I'd probably put it at the top.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/Directory.java
FinlayRJW: ~~I'm 50//50 on this is it universally true that any `FileFactory` would want an `ensureExists` - I'm questioning the usage of `ensureExists` at all now! I can't think of a situation that would call for creating a directory and not creating a file to go in it afterwards~~ 

Actually we probably do want `ensureExists` on `FileFactory` as there are other directories `BuildDir` etc. Also then we can remove `Directory` and just use `ArbitrarySrcDir`

FinlayRJW: I suppose the other option is to make a root `Directory` which extends `FileFactory` then all other dirs extend that but that feels over complicated

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
```diff
@@ -17,19 +17,36 @@
...
 import com.palantir.gradle.testing.files.gradle.GradleFile;
 import com.palantir.gradle.testing.files.gradle.RegularGradleFile;
 import com.palantir.gradle.testing.files.properties.PropertiesFile;
 import com.palantir.gradle.testing.files.yaml.YamlFile;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 
 public interface FileFactory {
```
CRogers: Let's rename `FileFactory` to `Directory` - I think it's clearer

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
```diff
@@ -17,19 +17,36 @@
...
     Path path();
 
+    default FileFactory ensureExists() {
```
crogoz: nit - the method name sounds to me a bit more like a Precondition check, rather then the creation of the directory structure

crogoz (commented):
maybe the PR description should be updated to change Directory -> FileFactory and remove the `Automatically creates directories on instantiation`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
FinlayRJW: updated to `mkdirs`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/FileFactory.java
FinlayRJW: I've updated     
```java
default Directory directory(String path) {
        return new ArbitrarySrcDir(resolvePath(path));
}
```
not sure if this is bad as wont get anything specific on ArbitrarySrcDir but felt odd to return ArbitrarySrcDir and not Directory

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/Directory.java
```diff
@@ -17,19 +17,36 @@
...
+        }
+        return this;
+    }
+
     default ArbitraryFile file(String path) {
         return new ArbitraryFile(resolvePath(path));
     }
 
+    default Directory directory(String path) {
+        return new ArbitrarySrcDir(resolvePath(path));
```
CRogers: Let's call this `ArbitraryDirectory` while we can still easily rename it.

CRogers: To go with `ArbitraryFile` and because this won't always be a source dir

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/Directory.java
```diff
@@ -17,19 +17,36 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 
-public interface FileFactory {
+public interface Directory {
     Path path();
 
+    default Directory mkdirs() {
```
CRogers: Never really been a fan of `mkdirs` - can we call this `ensureExists()` or something more descriptive?

FinlayRJW: from https://github.com/palantir/gradle-plugin-testing/pull/234#discussion_r2460488847 maybe createDirectories?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/GradleSourceSet.java
```diff
@@ -30,7 +30,7 @@ public JavaSrcDir java() {
         return new JavaSrcDir(path.resolve("java"));
     }
 
-    public ArbitrarySrcDir srcDir(String srcDirName) {
-        return new ArbitrarySrcDir(this.path.resolve(srcDirName));
+    public ArbitraryDirectory srcDir(String srcDirName) {
```
CRogers: `Directory` rather than `ArbitraryDirectory`

CRogers: Other than that looks good

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/arbitrary/ArbitraryDirectory.java
```diff
@@ -18,10 +18,10 @@
 
 import com.google.errorprone.annotations.RestrictedApi;
 import com.palantir.gradle.testing.RestrictedCreation;
-import com.palantir.gradle.testing.files.FileFactory;
+import com.palantir.gradle.testing.files.Directory;
 import java.nio.file.Path;
 
-public record ArbitrarySrcDir(Path path) implements FileFactory {
+public record ArbitraryDirectory(Path path) implements Directory {
```
felixdesouza, line 24: separate to the spirit of this change, it's *not* additive, this is a deletion and an addition to be clear, please update the description

felixdesouza, line 24: I'm not sure how I feel about `ArbitrarySrcDir` being collapsed into a `ArbitraryDirectory`, feels like having the hierarchy:

```
Directory
  |
ArbitraryDirectory
  |
ArbitrarySrcDir
```

But not fully fleshed out what I would want to live on `ArbitrarySrcDir`

felixdesouza: I see @CRogers has mostly talked about it [here](https://github.com/palantir/gradle-plugin-testing/pull/234#discussion_r2461155686), let's just see how it goes!

FinlayRJW: sorry - PR description became out of date - will fix!

CRogers (approved):

PR #235, opened 2025-10-23
Add custom string formatting

FinlayRJW:
## Before this PR
  

Using `.formatted()` breaks IDE language injection - the IDE cannot provide syntax highlighting for the string content (e.g., Gradle DSL in build files).

  ## After this PR

Screenshot 2025-10-24 at 10 17 21

  

  ### File Formatting Support (files package)
  All `ProjectFile` implementations now support Java's `String.formatted()` syntax via varargs, **preserving IDE language injection and syntax highlighting**:

  **ProjectFile.java** - Base formatting methods
  - `overwrite(String text, Object... args)`
  - `append(String text, Object... args)`
  - `appendLine(String line, Object... args)`
  - `prepend(String text, Object... args)`
  - `prependLine(String line, Object... args)`

  **Type-specific files with @Language annotations preserved**
  - **GradleFile.java** - All methods with `@Language("Gradle")` + `@FormatString`
  - **JavaFile.java** - All methods with `@Language("Java")` + `@FormatString`
  - **PropertiesFile.java** - All methods with `@Language("Properties")` + `@FormatString`
  - **YamlFile.java** - All methods with `@Language("YAML")` + `@FormatString`
  - **JavaSrcDir.java** - `writeClass()` supports formatting
  - **SettingsGradleFile.java** - Updated to use new formatting methods

  **ProjectFileFormattingTest.java** - Tests demonstrating both approaches 

work, with a test explicitly documenting the IDE benefit

  **Example usage:**
```java
  // Before - loses syntax highlighting after .formatted()
  String taskName = "myCustomTask";
      tasks.register('%s') {
          doLast {
              println 'Hello'
          }
      }
      """.formatted(taskName));  // No Gradle syntax highlighting in IDE

  // After - preserves syntax highlighting with varargs
  String taskName = "myCustomTask";
      tasks.register('%s') {
          doLast {
              println 'Hello'
          }
      }
      """, taskName);  // Full Gradle syntax highlighting in IDE
```

  ==COMMIT_MSG==
  Add custom formatting
  - Add @FormatMethod overloads to ProjectFile preserving IDE language injection
 ==COMMIT_MSG==

  ## Possible downsides?
  

  None expected. All changes are additive


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
```diff
@@ -16,23 +16,44 @@
...
 
+import com.google.errorprone.annotations.FormatMethod;
+import com.google.errorprone.annotations.FormatString;
 import com.palantir.gradle.testing.files.ProjectFile;
 import org.intellij.lang.annotations.Language;
 
 public interface GradleFile extends ProjectFile<GradleFile> {
+    @Override
+    @FormatMethod
+    default GradleFile overwrite(@Language("Gradle") @FormatString String text, Object... args) {
```
CRogers: Does these errorprone `@FormatString`/`@FormatMethod` help intellij understand it's a format string and do something sensible (like highlight the wrong number of args) or just helps error-prone?

I just looked in the `org.jetbrains:annotations` and there is a `@PrintFormat` annotation in there as well? Maybe we want both? I'd be interested what the effect of having `@PrintFormat` _and_ `@Language` is?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/java/JavaFile.java
```diff
@@ -26,19 +28,38 @@ public record JavaFile(Path path) implements ProjectFile<JavaFile> {
...
     @Override
     public JavaFile append(@Language("Java") String text) {
         return ProjectFile.super.append(text);
     }
 
     @Override
-    public JavaFile appendLine(@Language("Java") String line) {
-        return ProjectFile.super.appendLine(line);
+    @FormatMethod
+    public JavaFile appendLine(@Language("Java") @FormatString String line, Object... args) {
```
CRogers: Why no non-format-string for `appendLine`?

CRogers: (and `prependLine`)?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/java/JavaSrcDir.java
```diff
@@ -32,7 +33,8 @@ public record JavaSrcDir(Path srcDirPath) {
     @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
     public JavaSrcDir {}
 
-    public JavaFile writeClass(@Language("Java") String javaSource) {
+    @FormatMethod
+    public JavaFile writeClass(@Language("Java") String javaSource, Object... args) {
```
CRogers: Also question whether we need a separate format string vs non-format string here, or we can just have everything (ie all the method in `ProjectFile`) use the format string variant

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
CRogers: Would be great to have some screenshots of this on the PR to see the difference too

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/java/JavaSrcDir.java
FinlayRJW: yeah we probably need to add one -else you can run into error-prone complaints if you try and format yourself

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
FinlayRJW: `@FormatString`/`@FormatMethod` seem only linked to error-prone doesn't seem to effect the highlighting - I've not played around with `@PrintFormat` will investigate now

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/java/JavaFile.java
FinlayRJW: whoops must have missed them!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
```diff
@@ -35,31 +37,60 @@ default T overwrite(String text) {
         return (T) this;
     }
 
+    @FormatMethod
+    default T overwrite(@FormatString String text, Object... args) {
+        return overwrite(args.length > 0 ? text.formatted(args) : text);
+    }
+
     default T append(String text) {
```
CRogers, line 45: Ideally we'd also make an errorprone to check that people don't do:

```java
file.append("foo %s".formatted(3))
file.append(String.format("foo %s", 3))
```

...and force them to do:

```java
file.append("foo %s", 3)
```

...instead to ensure people get the syntax highlighting. But I need to find the time to get #221 in first, so later PR.

FinlayRJW: interesting - I've only had to leave all the annoying non-formatted methods around to cover cases where users pass in `file.append(String.format("foo %s", 3))` as without this you get a hard to debug error-prone message 

```
/Volumes/git/gradle-plugin-testing/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java:52: error: [FormatStringAnnotation] Local format string variables must only be assigned to compile time constant values. Invalid format string assignment: @Language(value = "Gradle")
        appendLine(includeLine);
                  ^
  String includeLine = "include \'%s\'".formatted(projectPath)
    (see https://errorprone.info/bugpattern/FormatStringAnnotation)
```

but if we are going to add an errorprone shortly anyway maybe we can just live with this for now and remove all these old methods?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
FinlayRJW: okay so `@PrintFormat` is so the IDE know we are doing formatting and tells the user errors i.e. `myprintf("%d\n", "hello"); // warning: a number expected instead of "hello"`

FinlayRJW: So I think we need both as @FormatString/@FormatMethod is purely error prone

FinlayRJW: added @PrintFormat so we get formatting errors in IDE

CRogers: It's crazy IntelliJ support external `@PrintFormat` annotations but not external `@Language`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/java/JavaSrcDir.java
```diff
@@ -45,6 +47,11 @@ public JavaFile writeClass(@Language("Java") String javaSource) {
         return fileByClassName(canonicalClassName).overwrite(javaSource);
     }
 
+    @FormatMethod
+    public JavaFile writeClass(@Language("Java") @PrintFormat String javaSource, Object... args) {
```
CRogers: Needs a `@FormatString` too?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
CRogers: I think we need to keep the old methods so people _can_ pass specially calculated/non-constant values when not using format strings.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
```diff
@@ -35,31 +37,60 @@ default T overwrite(String text) {
...
     }
 
+    @FormatMethod
+    default T prependLine(@FormatString String line, Object... args) {
+        return prependLine(format(line, args));
+    }
+
+    @FormatMethod
+    private static String format(String text, Object... args) {
+        return args.length > 0 ? text.formatted(args) : text;
```
CRogers: Do we need to do this, or can all of these methods just call `text.formatted(args)`, even when `args` is empty?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
FinlayRJW: do they! I hadn't tested yet seems crazy!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
FinlayRJW: I think that should be safe now we have all the non-format methods - I was slightly worried this might have caused issues if you happend to have a string with %s but didnt want formatting

CRogers: that should be fine, I think this can be inlined

CRogers (approved):

PR #236, opened 2025-10-23
Add `addDependency` method to `GradleFile`

FinlayRJW:
## Before this PR
Had no way to easily add dependencies to gradle files (a common operation)
## After this PR
Now a user can do 

```java
        rootProject
                .buildGradle()
                .addDependency("implementation", "com.google.guava:guava:31.1-jre")
                .addDependency("testImplementation", "junit:junit:4.13.2");
```

==COMMIT_MSG==
Add `addDependency` method to `GradleFile`
==COMMIT_MSG==

## Possible downsides?
may be adding a rarely used method and confusing the API?

### DIFF
--- a/gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/gradle/GradleFileTest.java
+++ b/gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/gradle/GradleFileTest.java
@@ -26,7 +26,7 @@ import org.junit.jupiter.api.Test;
 class GradleFileTest {
     @Test
     void addDependencies_single_dependency_with_default_configuration(RootProject rootProject) {
-        rootProject.buildGradle().addDependencies("com.google.guava:guava:31.1-jre");
+        rootProject.buildGradle().addDependency("implementation", "com.google.guava:guava:31.1-jre");
 
         assertThat(rootProject.buildGradle().text()).contains("""
             dependencies {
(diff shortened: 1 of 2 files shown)


gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/gradle/GradleFileTest.java
```diff
@@ -0,0 +1,93 @@
...
+                .addDependency("implementation", "com.google.guava:guava:31.1-jre")
+                .addDependency("testImplementation", "junit:junit:4.13.2");
+
+        assertThat(rootProject.buildGradle().text()).contains("""
+            dependencies {
+                implementation 'com.google.guava:guava:31.1-jre'
+            }
+            dependencies {
+                testImplementation 'junit:junit:4.13.2'
+            }
```
CRogers, line 76: I guess we want to decide whether we should be clever and try to insert into existing dependency blocks or not.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/gradle/GradleFileTest.java
```diff
@@ -0,0 +1,93 @@
...
+            """);
+    }
+
+    @Test
+    void addDependencies_can_be_chained(RootProject rootProject) {
+        rootProject
+                .buildGradle()
+                .addDependency("implementation", "com.google.guava:guava:31.1-jre")
+                .addDependency("testImplementation", "junit:junit:4.13.2");
```
CRogers, line 68: I wonder whether this is actually more readable that just using the string formatting from #235. At least when you write that out manually, you can see exactly what the change to the gradle file is going to be.

FinlayRJW: yeah I'm not convinced this actually brings much value vs just writing the buildGradle think it might honestly just confuse people more than anything

CRogers: The only real benefit I could see of an `addDependency` method is if we want to get the dependency version from somewhere that is automatically updated, so the tests are always testing against latest. Might be worth putting a pin in this until we need that feature.

FinlayRJW: will close for now then can always re-open if we go that route

PR #238, opened 2025-10-24
rootProject.name is appended in settings file rather than prepended and replace insitu

FinlayRJW:
## Before this PR
If `rootProject.name` was not present i.e. during `beforeEach` it would be prepended. This would result in errors if you say added a `plugins` block in the `beforeEach` as `rootProject.name` would be added first

## After this PR
Now we first try and in-situ replace the all occurrence of `rootProject.name`. If `rootProject.name` is not present we append rather than prepend 

Now `rootProject.name` is appended

==COMMIT_MSG==
`rootProject.name` is appended in settings file rather than prepended 
==COMMIT_MSG==

## Possible downsides?
existing tests that checked the raw settings file content will be broken although I don't think any exist


gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -68,6 +70,7 @@ void sub_project_parameter_ending_in_project(GradleInvoker gradle, SubProject as
         assertThat(gradle.withArgs().buildsSuccessfully().output()).contains("hello from :asset");
     }
 
+    @Test
```
FinlayRJW, line 75: I assume this missing was a mistake?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
```diff
@@ -16,27 +16,25 @@
...
-                    .splitToStream(text)
-                    .filter(line -> !line.startsWith("rootProject.name"))
-                    .collect(Collectors.joining("\n"));
-        });
-
-        prependLine("rootProject.name = '%s'".formatted(rootProjectName));
+        String newLine = "rootProject.name = '%s'".formatted(rootProjectName);
+        String placeholder = "<<<ROOTPROJECT_PLACEHOLDER>>>";
+
+        edit(text -> text.contains("rootProject.name")
```
FinlayRJW: If we contain `rootProject.name` replace the first with a place holder, remove the rest then convert the place holder to the new line. This lets us replace insitu and de-duplicate. If not present we append

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFileTest.java
```diff
@@ -54,7 +54,48 @@ void maintains_ending_newlines() {
 
             settingsGradleFile.rootProjectName("name");
 
-            settingsGradleFile.assertThat().hasContent("rootProject.name = 'name'\nsomething already here\n\n");
+            settingsGradleFile.assertThat().hasContent("something already here\n\nrootProject.name = 'name'\n");
```
FinlayRJW, line 59: now we append

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
FinlayRJW: not sure if this is overkill and we could just do a replace all?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
```diff
@@ -16,27 +16,22 @@
...
     public SettingsGradleFile rootProjectName(String rootProjectName) {
-        edit(text -> {
-            return Splitter.on('\n')
-                    .splitToStream(text)
-                    .filter(line -> !line.startsWith("rootProject.name"))
-                    .collect(Collectors.joining("\n"));
-        });
-
-        prependLine("rootProject.name = '%s'".formatted(rootProjectName));
+        String newLine = "rootProject.name = '%s'".formatted(rootProjectName);
```
felixdesouza: rename this to better reflect what this is, I thought `newLine` was literally a new line.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFileTest.java
```diff
@@ -54,7 +54,49 @@ void maintains_ending_newlines() {
...
+                """);
+        }
+
+        @Test
+        void replaces_all_instances() {
+            settingsGradleFile.overwrite("""
+                // before
+                rootProject.name = 'init'
+                // during
+                rootProject.name = 'second'
```
felixdesouza, line 89: this is weird, there's no reason for you to set the name of the `rootProject` multiple times. If your settings plugin code is relying on that behaviour, you're not going to have a good time. 

i.e.

```
rootProject.name = 'init'

// do something that depends on the name being `init`

rootProject.name = 'what-i-actually-want'
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
```diff
@@ -16,27 +16,22 @@
...
+
+        edit(text -> text.contains("rootProject.name")
```
felixdesouza: separately to this PR, this editing workflow seems error prone - as this PR is demonstrating by fixing a previous error).

It kinda feels like the `SettingsGradleFile` should be some sort of builder of state because all the things it comprises are structural. 

e.g. keeping hold of the state `rootProjectName`, having a list of "blocks" for `plugins`, being able to add arbitrary blocks that have a particular place where it lives.

Then when it comes to starting a build, or invoking a method that looks at the contents, it renders the file like a template i.e.

```
rootProject.name = '#rootProjectName'

#pluginsBlock

#extraBlocks
```

`edit` can edit things in the correct place after the required scaffolding has been set up. Otherwise it's just a regex mess imo.

Does not need to block this PR, but we should definitely think about it. It also plays nicely into having good helpers in `RegularGradleFile` e.g. `addDependencies`, we can then just add something like `addBlocks` or something. But I feel like having to edit the whole file in this way is somewhat ghastly.

Let's make an issue that we can discuss next week

FinlayRJW: yeah I think we want something like this, i.e. something that respects the order of `buildScript`, `plugins` block, everything else, otherwise it can be easy to break the build.gradle file

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFileTest.java
FinlayRJW: Cool now we throw if there are multiple `rootProject.name` blocks

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
CRogers: ??

CRogers (approved):

PR #244, opened 2025-10-27
Errorprone to force use of our formatting methods

FinlayRJW:
## Before this PR
We have a bunch of methods which take formatting args currently we have no way to force users to use these methods 
## After this PR
Now we have an errorprone to enforce usage - this ensures users always get code highlighting

==COMMIT_MSG==
Errorprone to force use of our formatting methods
==COMMIT_MSG==


gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestStringFormatting.java
```diff
@@ -0,0 +1,93 @@
...
+        List<? extends ExpressionTree> arguments = tree.getArguments();
+        if (arguments.isEmpty()) {
+            return Description.NO_MATCH;
+        }
+
+        ExpressionTree firstArg = arguments.get(0);
+
+        if (FORMATTED_STRING.matches(firstArg, state)) {
+            return describeMatch(tree);
+        }
```
CRogers, line 87: We might introduce methods on `ProjectFile` / `JavaSrcDir` that take a `String` and it's ok for it to be formatted manually and this will pick it up. To make it a bit more futureproof, could we check if there's a method with the same name that is marked with `@FormatMethod` with an arg marked `@FormatString`?

CRogers: Considering we're probably going to introduce more things like `JavaSrcDir#writeClass`, it might make sense to make on all methods that are from this library where `String.format` is used to (and see if it has a `@FormatMethod` equivalent). At the cost of a little more complication time, but no need to keep the `TARGET_METHOD` matcher updated. Gradle plugin repos are normally fast anyway.

FinlayRJW: good catch - currently `javaDir.fileByClassName(` get errorproned even through they dont have a formatted overload!

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestHelpers.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.palantir.gradle.testing.errorprone;
+
+import com.google.errorprone.VisitorState;
+import com.google.errorprone.matchers.Matcher;
+import com.google.errorprone.matchers.Matchers;
+import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.Tree;
+
+public final class GradleTestHelpers {
```
felixdesouza: nit: `GradleTestHelpers` -> `GradlePluginTestHelpers`

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestStringFormatting.java
```diff
@@ -0,0 +1,129 @@
...
+        if (firstArg instanceof IdentifierTree identifier) {
+            if (isIdentifierInitialisedWithFormattedString(identifier, state)) {
+                return describeMatch(tree);
+            }
+        }
+
+        return Description.NO_MATCH;
+    }
+
+    private static boolean isLibraryMethod(MethodInvocationTree tree, VisitorState state) {
```
felixdesouza: should this move to `GradlePluginTestHelpers` ?

gradle-plugin-testing-error-prone/src/test/java/com/palantir/gradle/testing/errorprone/GradleTestStringFormattingTest.java
```diff
@@ -0,0 +1,244 @@
...
+                    // BUG: Diagnostic contains: GradleTestStringFormatting
+                    file.append(String.format("foo %s", 3));
+                    // BUG: Diagnostic contains: GradleTestStringFormatting
+                    file.appendLine(String.format("foo %s", 3));
+                    // BUG: Diagnostic contains: GradleTestStringFormatting
+                    file.prepend(String.format("foo %s", 3));
+                    // BUG: Diagnostic contains: GradleTestStringFormatting
+                    file.prependLine(String.format("foo %s", 3));
+                    // BUG: Diagnostic contains: GradleTestStringFormatting
+                    file.overwrite(String.format("foo %s", 3));
```
felixdesouza, line 53: these are all fixable no?

I wonder if we get fixes in, whether the error message you get when you run `compileJava` will be enough for the agent to run errorprone fixes without trying to fix it itself (which may be errorprone lol)

FinlayRJW: I didn't add an autofix purely as everything should already be perfect - but if we think that the error messages might improve agents can add!

FinlayRJW: added autofix for these but not this style

```java
                    String str = "hi %s".formatted("world");
                    // BUG: Diagnostic contains: GradleTestStringFormatting
                    file.append(str);
```

as I don't know what else the user is doing with the string

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
```diff
@@ -0,0 +1,49 @@
...
+            Matchers.isInstance(ClassTree.class),
+            Matchers.hasAnnotation("com.palantir.gradle.testing.junit.GradlePluginTests")));
+
+    private static final String LIBRARY_PACKAGE = "com.palantir.gradle.testing.";
+
+    static boolean notWithinGradlePluginTests(Tree tree, VisitorState state) {
+        return !WITHIN_GRADLE_PLUGIN_TESTS_CLASS.matches(tree, state);
+    }
+
+    static boolean isLibraryMethod(MethodInvocationTree tree, VisitorState state) {
```
CRogers: Let's call this `isGradlePluginTestsLibraryMethod` - library method is too vague.

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestStringFormatting.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private static final Matcher<ExpressionTree> FORMATTED_STRING = Matchers.anyOf(INSTANCE_FORMATTED, STRING_FORMAT);
+
+    @Override
+    public Description matchMethodInvocation(MethodInvocationTree tree, VisitorState state) {
+        if (GradlePluginTestHelpers.notWithinGradlePluginTests(tree, state)) {
+            return Description.NO_MATCH;
+        }
+
+        if (!GradlePluginTestHelpers.isLibraryMethod(tree, state)) {
```
CRogers: I'd put this method first as it's going to be slightly faster to check.

CRogers (commented):
Looks good except for comments!

CRogers (approved):

PR #245, opened 2025-10-27
Add support for running GradlePluginTests with configuration-cache

crogoz:
## Before this PR

Groovy Test spec that runs with [Configuration Cache enabled](https://github.com/palantir/gradle-plugin-testing/blob/develop/configuration-cache-spec/src/main/groovy/com/palantir/gradle/plugintesting/ConfigurationCacheSpec.groovy#L23).

## After this PR

Ability to run GradlePluginTests with "--configuration-cache" enabled when the extension properrty `gradleTestUtils.isConfigurationCacheEnabled` is set.

Based on the value of the configuration cache, we will instantiate a `DefaultGradleInvoker` or a `ConfigurationCacheInvoker`. 

The `ConfigurationCacheInvoker` will do: 
1. run the tasks with `--configuration-cache` added to the list of arguments. It will throw an `UnexpectedConfigurationCacheFailure` exception if the configuration cache was not stored/failed. 
2. run the tasks above including `--configuration-cache` & `--dry-run` to make sure the configuration cache was reused. It will throw an `UnexpectedConfigurationCacheFailure` if not. 

Added a new `DisabledConfigurationCache` which will disable per test class/method the configuration-cache. This will be useful later when we are migrating the GradlePluginTests to use by default the configuration-cache and we will need to disable the tests that are incompatible with cc.

==COMMIT_MSG==
Ability to run GradlePluginTests with configuration-cache.
==COMMIT_MSG==


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvocation.java
```diff
@@ -32,10 +40,60 @@ public GradleInvocation withEnvironment(Map<String, String> environment) {
...
+            GradleRunner configurationCacheRunner = GradleRunner.create()
+                    .withProjectDir(gradleRunner.getProjectDir())
+                    .withDebug(gradleRunner.isDebug())
+                    .forwardOutput()
+                    .withGradleVersion(gradleVersion.version())
+                    .withPluginClasspath(gradleRunner.getPluginClasspath())
+                    .withEnvironment(gradleRunner.getEnvironment())
+                    .withArguments(ImmutableList.<String>builder()
+                            .addAll(gradleRunner.getArguments())
+                            .add("--dry-run")
```
crogoz: for config-cache to be reused, we need to run the same tasks. In order to avoid running the same tasks twice, we can pass a `--dry-run`. 
Tested that config cache is triggered by a dry-run: 

```
Reusing configuration cache.
:compileJava SKIPPED
:help SKIPPED

BUILD SUCCESSFUL in 62ms
Configuration cache entry reused.
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvocation.java
```diff
@@ -32,10 +40,60 @@ public GradleInvocation withEnvironment(Map<String, String> environment) {
...
+                            .build());
+            InvocationResult configurationCacheResult = new InvocationResult(configurationCacheRunner.build());
+            assertConfigCacheReused(configurationCacheResult.output());
+        }
+
+        return result;
```
crogoz: always returning the result of the first invocation.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvocation.java
```diff
@@ -16,14 +16,22 @@
...
+
+import com.google.common.collect.ImmutableList;
+import java.io.File;
 import java.util.Map;
 import org.gradle.testkit.runner.GradleRunner;
 
 public final class GradleInvocation {
     private final GradleRunner gradleRunner;
+    private final GradleVersion gradleVersion;
+    private final boolean configurationCacheEnabled;
```
CRogers: Rather than taking a boolean and doing if statements, can we instead try and change `GradleInvoker` and `GradleInvocation` into interfaces, and have a `BasicGradleInvoker` and a `ConfigurationCacheGradleInvoker` (likely the same for `GradleInvocation` too). The configuration cache versions should just be able to call the basic versions as many times as they need - we might be able to avoid copying the code for creating the `GradleRunner` again on line 50 too.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTest.java
```diff
@@ -0,0 +1,50 @@
...
+import com.palantir.gradle.testing.execution.TaskOutcome;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+@WithConfigurationCache
+final class ConfigurationCacheTest {
+
+    @Test
+    void testConfigurationCache(GradleInvoker invoker, RootProject rootProject) {
```
CRogers: Test methods should not start with `test`, it's superfluous. Additionally, test names should be a `snake_case_english_sentence`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvocation.java
```diff
@@ -32,10 +40,60 @@ public GradleInvocation withEnvironment(Map<String, String> environment) {
...
     public InvocationResult buildsSuccessfully() {
-        return new InvocationResult(gradleRunner.build());
+        InvocationResult result = new InvocationResult(gradleRunner.build());
+
+        if (configurationCacheEnabled) {
+            assertConfigCacheStored(result.output());
+
+            // Run a second time to verify the cache is reused, we don't need to run the tasks --dry-run is enough to
+            // check if the configuration cache was used.
+            GradleRunner configurationCacheRunner = GradleRunner.create()
```
FinlayRJW: I've seen cases where the second run doesn't reuse the cache but the subsequent do - normally related to file generation. I'm not sure if that behaviour is ever valid so the solution might just be that we want to fail those cases but could also be nice to have some kind of ratchet - although in those edge cases we could just get users to write there CC tests

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTest.java
```diff
@@ -0,0 +1,50 @@
...
+        rootProject.buildGradle().append("plugins { id 'java' }");
+
+        rootProject.mainSourceSet().java().writeClass("""
+                package com.palantir.example;
+
+                record MyClass() {}
+            """);
+
+        InvocationResult result = invoker.withArgs("compileJava", "help").buildsSuccessfully();
```
CRogers: I'm not sure what this is currently testing?

In terms of tests, I think we could have at least a few tests:

* In the build.gradle, print whether the configuration cache is enabled and check it is disabled when `@WithConfigurationCache` is not present, and is enabled when present. You can use the [start parameter](https://docs.gradle.org/9.1.0/javadoc/org/gradle/StartParameter.html#isConfigurationCacheRequested()) / [configuration cache build feature](https://docs.gradle.org/9.1.0/javadoc/org/gradle/api/configuration/BuildFeatures.html#getConfigurationCache()). This actually raises a good questions - we might not want to run with configuration on Gradle 7, as the implementation and error messages are bad.
* Fails when there is a basic configuration cache error eg executing a process at configuration time
* Fails when there is an issue storing the cache entry
* Fails when there is an issue loading the cache entry
* Succeeds when there are no configuration cache issues

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvocation.java
```diff
@@ -32,10 +40,60 @@ public GradleInvocation withEnvironment(Map<String, String> environment) {
...
+                .as(String.format(
+                        "Running @WithConfigurationCache: Expected the configuration-cache %s directory to exits.",
+                        configurationCacheDir));
+    }
+
+    private void assertConfigCacheReused(String output) {
+        assertThat(output.contains("Configuration cache entry reused."))
+                .as(String.format(
+                        "Running @WithConfigurationCache: Expected configuration cache entry to be reused, but it"
+                                + " wasn't. Output: %s",
```
CRogers: I think, especially for the second invocation, we need a really good error message that is very explicit in what happened. We need to tell the dev that first invocation worked and stored the cache entry, but then we ran again with `--dry-run` to verify that loading the cache still worked and this failed, and how they should diagnose it (look at the report? build scan? can we parse these from the output and put it in the error message?)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvocation.java
```diff
@@ -32,10 +40,60 @@ public GradleInvocation withEnvironment(Map<String, String> environment) {
...
+        assertThat(output.contains("Configuration cache entry stored."))
+                .as(String.format(
+                        "Running @WithConfigurationCache: Expected configuration cache entry to be stored, but it"
+                                + " wasn't. Output: %s",
+                        output));
+
+        File configurationCacheDir = new File(gradleRunner.getProjectDir(), ".gradle/configuration-cache");
+        assertThat(configurationCacheDir.exists())
+                .as(String.format(
+                        "Running @WithConfigurationCache: Expected the configuration-cache %s directory to exits.",
```
CRogers: s/exits/exist

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvocation.java
FinlayRJW: I'm pretty sure that was the case here https://pl.ntr/2xX - but its possible I also just didn't fix it properly!

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTest.java
CRogers: I actually think we should make an errorprone check for this, to ensure that people do this for our Gradle plugins tests.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WithConfigurationCache.java
```diff
@@ -0,0 +1,35 @@
...
+ * Annotation that can be used alongside {@link GradlePluginTests} to enable configuration cache testing.
+ * When applied, all Gradle tasks executed via {@link com.palantir.gradle.testing.execution.GradleInvoker} will run with
+ * the "--configuration-cache" flag. For successful builds ({@link com.palantir.gradle.testing.execution.GradleInvocation#buildsSuccessfully}),
+ * a second dry-run is automatically performed to verify that the configuration cache is properly reused.
+ * The {@link com.palantir.gradle.testing.execution.InvocationResult} returned always represents the outcome of the first run.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(ConfigurationCacheExtension.class)
+public @interface WithConfigurationCache {}
```
CRogers: In terms of actually rolling this out, not sure how I feel about opting in (like this with `@WithConfigurationCache`) vs a global opt in at the Gradle extesion level and opting out with `@ConfigurationCacheDisabled`. Or even further - default having configuration cache on without any Gradle configuration and relying on a disable annotation.

My worry with the opting in is people will just forget to add it to new tests and we want to ensure all new code/test code is configuration cache compatible.

But in order for upgrades to not get blocked with opt out, we need to work out a way to feed back test failures to the excavator which will enable it to add `@ConfigurationCacheDisabled` onto the relevant tests/test classes. Opt-in, it's a little easier as you can make a parameterised excavator that tries to opt in each test class, then the excavator will just not merge if it doesn't work.

crogoz: My thinking was to add this flag to all tests that are extending [ConfigurationCacheSpec](https://github.com/palantir/gradle-plugin-testing/blob/develop/configuration-cache-spec/src/main/groovy/com/palantir/gradle/plugintesting/ConfigurationCacheSpec.groovy#L23) example [here](https://github.com/palantir/gradle-git-version/blob/25e62d5968610a2374006bd307358bb4574c4d00/src/test/groovy/com/palantir/gradle/ConfigurationCacheTest.groovy#L21). This way we can do the migration of the nebula tests -> new java framework 1-1. 

Then I think it would be pretty easy to add an extension similar to how we set the Gradle versions and make it by default using the CC and add a workflow to add the `@ConfigurationCacheDisabled`. But I am thinking that would be a separate migration from the nebula test -> new Java framework. wdyt ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private static boolean isJavaDebugAgentLoaded() {
+        // When you run a test with debug in intellij, it passes an arg to the test process like:
+        //   -agentlib:jdwp=transport=dt_socket,server=n,suspend=y,address=127.0.0.1:54342
+        // We can use this to detect whether we should run with Gradle tooling `.withDebug` or not,
+        // `withDebug(true)` will run the Gradle tooling inside the same JVM as the test, meaning
+        // debugging works, whereas `withDebug(false)` will run Gradle in a new daemon. There can be
+        // differences between these two modes!
+        return ManagementFactory.getRuntimeMXBean().getInputArguments().stream()
```
crogoz, line 59: FLUP: cc is not compatible with testkit's debugging option see issue: gradle/issues/25846

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -35,8 +36,9 @@ void root_project_parameter(GradleInvoker gradle, RootProject rootProject) {
             println "project name: ${name}"
             """);
 
-        assertThat(gradle.withArgs().buildsSuccessfully().output()).contains("hello from :");
-        assertThat(gradle.withArgs().buildsSuccessfully().output()).contains("project name: root");
+        InvocationResult invocation = gradle.withArgs().buildsSuccessfully();
+        assertThat(invocation.output()).contains("hello from :");
+        assertThat(invocation.output()).contains("project name: root");
```
crogoz: I've enabled cc here. Previously we were running the tasks twice, which means that cc will not run the second time - hence the second run will not contain the expected output lines. Same below.

gradle-plugin-testing-junit/build.gradle
```diff
@@ -30,6 +30,11 @@ dependencies {
     testFixturesImplementation 'org.junit.jupiter:junit-jupiter-params'
 }
 
+tasks.named('test', Test) {
+    // by default the tests should run with configuration cache enabled.
+    systemProperty 'com.palantir.gradle.testing.configuration_cache_enabled', 'true'
```
crogoz, line 35: Once we publish this version, I'll add the `isConfigurationCacheEnabled = true` in the `gradleTestutils` extension

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvoker.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package com.palantir.gradle.testing.execution;
+
+import com.google.common.collect.ImmutableList;
+import com.google.errorprone.annotations.RestrictedApi;
+import com.palantir.gradle.testing.RestrictedCreation;
+import java.nio.file.Path;
+
+public record ConfigurationCacheInvoker(Path projectDir, GradleInvoker gradleInvoker) implements GradleInvoker {
```
CRogers: We should make this a regular class so we can hide `.gradleInvoker()` etc

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvoker.java
```diff
@@ -0,0 +1,45 @@
...
+    public ConfigurationCacheInvoker {}
+
+    @Override
+    public GradleInvocation withArgs(String... args) {
+        String[] withConfigurationCacheEnabled = ImmutableList.<String>builder()
+                .add(args)
+                .add("--configuration-cache")
+                .build()
+                .toArray(String[]::new);
+        GradleInvocation initialGradleInvocation = gradleInvoker.withArgs(withConfigurationCacheEnabled);
```
CRogers, line 49: It might be nice to be able to chain `withArgs` eg `gradleInvoker.withArgs(args).withArgs("--configuration")` - not just for this case but for using the framework too

CRogers: Probably best left to a follow up PR

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
CRogers: Good catch, we should probably just disable configuration cache when debugging and paste a warning log line that we've done so in each test. This is [what we do in suppressible-error-prone](https://github.com/palantir/suppressible-error-prone/blob/691db97152d244687b4d39c8b11cdb7721c10b41/gradle-suppressible-error-prone/src/test/groovy/com/palantir/gradle/suppressibleerrorprone/SuppressibleErrorPronePluginIntegrationTest.groovy#L2027-L2039), would be great to get it in the framework directly.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheParameterExtension.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    public void beforeAll(ExtensionContext context) {
+        setConfigurationCache(context);
+    }
+
+    @Override
+    public void beforeEach(ExtensionContext context) {
+        setConfigurationCache(context);
+    }
```
CRogers, line 37: I wonder if it's safer to just lazily compute whether configuration cache is enabled or not in the store rather than setting them here - can't forget to set it in all the relevant locations then.

CRogers: I guess that's a bit strange, as you need to look at the enclosing elements to find configuration cache, and how would we handle a `@ConfigurationCacheDisabled` arg on a test when there's a `@WithConfigurationCache` on the test class...

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
CRogers: Oh, interesting! I have a feeling this more than anything may break existing tests that use printlns at configuration time a lot (though I guess they should be doing it within task actions if they were good).

I wonder if we wanted tests to be more or less configuration cache agnostic whether we delete the configuration cache dir after checking it can be reloaded with `--dry-run`. I'm not sure if it's a good idea, but it would still check configuration cache but run the configuration every time. Feels a little dodgy though, maybe we just keep it as an option in case converting existing tests is hard?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTests.java
```diff
@@ -0,0 +1,155 @@
...
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+class ConfigurationCacheTests {
+
+    @Test
+    void configuration_cache_is_enabled_by_default(GradleInvoker invoker, RootProject rootProject) {
+        rootProject.buildGradle().append("""
+            plugins { id 'java' }
+            def isConfigurationCacheRequested = services.get(BuildFeatures).configurationCache.requested.orElse(false).get()
```
CRogers: I'd make this a proper input to the task ([like this](https://github.com/palantir/gradle-baseline/blob/0e881a2145c6d0a1ad6ff9638d0702b8f3f34cb8/gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineModuleJvmArgsIntegrationTest.java#L127-L132)). Might as well right even test tasks "correctly" so we don't have to fix the tests when deprecations come around/Gradle becomes stricter on what tasks can do.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
```diff
@@ -0,0 +1,125 @@
...
+        assertConfigCacheStoredOrReused(result);
+        return result;
+    }
+
+    private void assertConfigCacheStoredOrReused(InvocationResult result) {
+        if (!result.output().contains("Configuration cache entry stored.")
+                // We might be still re-using the cache from a previous call.
+                && !result.output().contains("Configuration cache entry reused.")) {
+            throw new UnexpectedConfigurationCacheFailure(String.format("""
+                Build Execution failure caused by configuration cache issues. Expected configuration cache entry to be stored, but it wasn't.
```
CRogers, line 78: > ...to be stored, or reused from a previous invocation?

I guess the advantage of what I describe here about [here](https://github.com/palantir/gradle-plugin-testing/pull/245/files#r2481347198) about deleting the configuration cache dir after (before?) each call is that reusing instead of storing would never be a problem.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/GradleVersionsFromJunitParameterTest.java
```diff
@@ -35,6 +35,7 @@ void runs_tests_with_gradle_versions_from_junit_parameter() {
         EngineExecutionResults executionResults = EngineTestKit.engine("junit-jupiter")
                 .selectors(DiscoverySelectors.selectClass(GradleVersionsFromJunitParameterFixtureTest.class))
                 .configurationParameter("com.palantir.gradle.testing.gradle_versions_to_test", "7.6.5,8.14.3")
+                .configurationParameter("com.palantir.gradle.testing.configuration_cache_enabled", "false")
```
CRogers, line 38: We should add one of these engine test kit tests for this rather than just implicitly testing using the system prop on the test task.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheParameterExtension.java
crogoz: > how would we handle a @ConfigurationCacheDisabled arg on a test when there's a @WithConfigurationCache on the test class...

I removed this class in the latest commits. I only kept the `@DisableConfigurationCache`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
crogoz: ended up cleaning the config cache dir before each configuration cache call - https://github.com/palantir/gradle-plugin-testing/pull/245/files#diff-c9fc79e137dc208625f1d7fa8fc91d3360019b2b1bdb4067259fd301e8f82fc8R62

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/GradleVersionsFromJunitParameterTest.java
crogoz: added tests [here](https://github.com/palantir/gradle-plugin-testing/pull/245/files#diff-901888ff1f0efc7f87308e566960194f836a66df6c87b0dce2f83a64c4f6e2bd) for the cc configuration.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
```diff
@@ -0,0 +1,125 @@
...
+                        new InvocationResult(unexpectedBuildFailure.getBuildResult()));
+            }
+            throw unexpectedBuildFailure;
+        }
+    }
+
+    @Override
+    public InvocationResult buildsWithFailure() {
+        InvocationResult result = initialGradleInvocation.buildsWithFailure();
+        assertConfigCacheStored(result);
```
CRogers, line 69: What happens if it fails at configuration time, before executing any tasks? I think this is allowed by `buildsWithFailure` but the configuration cache entry won't be stored. Should have a test for it at least. Probably want to only check this if a task was executed/it was a task that failed.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
```diff
@@ -0,0 +1,125 @@
...
+        return result;
```
CRogers, line 72: I wonder if we actually want to check we can load configuration cache here if it was a task that failed. If the configuration was fine but a task failed, it should be fine to run with `--dry-run` (I think?)?

CRogers: Worth writing a test for.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvoker.java
```diff
@@ -0,0 +1,77 @@
...
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public ConfigurationCacheInvoker(Path rootProjectDir, GradleInvoker gradleInvoker) {
+        this.rootProjectDir = rootProjectDir;
+        this.gradleInvoker = gradleInvoker;
+    }
+
+    @Override
+    public GradleInvocation withArgs(String... args) {
+        // not reusing configuration-cache among multiple gradle invocations in a test.
+        cleanupConfigurationCache();
```
CRogers, line 42: Should we do this here or after a gradle invocation? I guess this might be a bit safer.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
crogoz: added tests [here](https://github.com/palantir/gradle-plugin-testing/pull/245/commits/663610fa8ca28f453b3425fbb52199b78ef33a9c)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
crogoz: added tests [here](https://github.com/palantir/gradle-plugin-testing/pull/245/commits/663610fa8ca28f453b3425fbb52199b78ef33a9c)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvoker.java
crogoz: I also think this would be safer, we would need in any case to ensure that when we start a new gradle invocation the configuration cache directory is not present anymore.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTests.java
```diff
@@ -0,0 +1,189 @@
...
+import org.gradle.testkit.runner.UnexpectedBuildSuccess;
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+class ConfigurationCacheTests {
+
+    @Test
+    void configuration_cache_is_enabled_by_default(GradleInvoker invoker, RootProject rootProject) {
+        rootProject.buildGradle().append("""
+            plugins { id 'java' }
```
FinlayRJW: nit: tests should use the new `plugins().add()`

FinlayRJW: nit: could we add the `java` plugin as part of beforeEach?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/WithoutConfigurationCacheTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+@DisabledConfigurationCache(reason = "testing class level annotation")
+class WithoutConfigurationCacheTest {
+
+    @Test
+    void configuration_cache_is_disabled_by_default(GradleInvoker invoker, RootProject rootProject) {
+        rootProject.buildGradle().append("""
+            plugins { id 'java' }
```
FinlayRJW: nit: tests should use the new plugins().add()

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/ConfigurationCacheFixtureTest.java
```diff
@@ -0,0 +1,51 @@
...
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+public class ConfigurationCacheFixtureTest {
+
+    @Test
+    void test_name(GradleInvoker gradleInvoker, RootProject rootProject) {
+        rootProject.buildGradle().append("""
+            plugins { id 'java' }
```
FinlayRJW: nit: tests should use the new plugins().add()

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTests.java
crogoz: changed in https://github.com/palantir/gradle-plugin-testing/pull/245/commits/b727bd7c720588fda75e0ef15ccdf703f8ade52f

FinlayRJW:
nit: now we have the docs PR in can we merge that in and update docs to include this!

FinlayRJW:
👍

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvoker.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.errorprone.annotations.RestrictedApi;
+import com.palantir.gradle.testing.RestrictedCreation;
+import java.io.File;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Path;
+import org.apache.commons.io.FileUtils;
+
+public final class ConfigurationCacheInvoker implements GradleInvoker {
```
pkoenig10, line 28: Classes like `ConfigurationCacheInvoker` and `ConfigurationCacheInvocation` should probably be package-private, right?

PR #250, opened 2025-10-30
Add `MavenRepo` test utility for publishing test dependencies

### LINKED ISSUE #248: Add `GradleDependencyGenerator#generateTestMavenRepo` equivalent to the testing framework
We often generate test maven repos so we should add this functionality to the new `junit` framework.

### Nebula implementation 
the nebula API works as follows:
```groovy
DependencyGraph dependencyGraph = new DependencyGraph(
                "a.b.c:c:1.0",
                "d.e.f:f:2.0 -> a.b.c:c:1.0|g.h.i:i:3.0");
DependencyGraph dependencyGraph = new DependencyGraph(graph)
GradleDependencyGenerator generator = new GradleDependencyGenerator(dependencyGraph, new File(projectDir, "build/testrepogen").toString())
generator.generateTestMavenRepo()
``` 

you can also use builder in nebula
```groovy
    DependencyGraphNode withDependency = new ModuleBuilder('group4:artifact4:4.4.4')
            .addDependency('group4a:artifact4a:44.44.44')
            .build()

    DependencyGraph dependencyGraph = new DependencyGraphBuilder()
            .addModule('group1:artifact1:1.1.1')
            .addModule('group2:artifact2:2.2.2')
            .addModule('group3:artifact3:3.3.3')
            .addModule(withDependency)
            .build()

GradleDependencyGenerator generator = new GradleDependencyGenerator(dependencyGraph, new File(projectDir, "build/testrepogen").toString())
```

Under the hood nebula creates a new gradle project, adds all the dependencies as subprojects then uses `publishMavenPublicationToMavenRepository` to get gradle to publish to the maven repo. 

### Proposed implementation

Create a new `MavenRepo` class that can be injected into the tests, it should have a `publish` method that can either take a `Module` or a similar string graph, this then creates a gradle project and publishes using `publishMavenPublicationToMavenRepository` the gradle project should be fully independent of the test 

## Usage Examples

```java
@GradlePluginTests
class MavenRepoExampleTests {

    @BeforeEach
    void setupCommonDependencies(MavenRepo repo, RootProject root) {
        // Shared dependencies available to all tests
        repo.publish("com.palantir:service-a:1.0.0");
        root.buildGradle().withMavenRepo(repo);
    }

    @Test
    void multiple_modules_with_dependencies(RootProject root, MavenRepo repo) {
        repo.publish(
            "com.external:library:1.0.0",
            "com.external:other:2.0.0 -> com.external:library:1.0.0|com.palantir:service-a:1.0.0"
        );
        
        root.buildGradle().append("""
            dependencies {
                implementation 'com.external:other:2.0.0'
            }
            """);
    }

    @Test
    void builder_pattern_with_custom_properties(RootProject root, MavenRepo repo) {
        Module lib = Module.of("com.external:library:1.0.0")
            .targetCompatibility(JavaVersion.VERSION_1_8)
            .build();

        Module app = Module.of("com.external:app:2.0.0")
            .dependsOn("com.external:library:1.0.0")
            .dependsOn("com.palantir:service-a:1.0.0")
            .targetCompatibility(JavaVersion.VERSION_1_8)
            .build();

        repo.publish(lib, app);
        
        root.buildGradle().append("""
            dependencies {
                implementation 'com.external:app:2.0.0'
            }
            """);
    }
```

I believe we should always only have a single `MavenRepo` instance i.e. the `MavenRepo` in beforeEach is the same as in the tests.

FinlayRJW:
## Before this PR
Testing Gradle plugins that depend on external Maven artifacts requires either:
- Using real external repositories (slow, unreliable, requires network access)
- Manually creating complex test repository setups
- Copying repository setup code across multiple test classes

This makes it difficult to write fast, isolated integration tests for Gradle plugins.

## After this PR
Tests can easily publish Maven artifacts to a local test repository using the new `MavenRepo` utility:

```java
@GradlePluginTests
class MyPluginTest {
    @BeforeEach
    void setupCommonDependencies(MavenRepo repo, RootProject root) {
        repo.publish(MavenArtifact.of("com.palantir:service-a:1.0.0"));
        root.buildGradle().withMavenRepo(repo);
    }
    
    @Test
    void test_with_dependencies(MavenRepo repo, GradleInvoker gradle) {
        repo.publish(
            MavenArtifact.of("com.external:library:1.0.0"),
            MavenArtifact.builder()
                .coordinate("com.external:other:2.0.0")
                .addDependency("com.external:library:1.0.0")
                .build()
        );
        
        gradle.withArgs("dependencies").buildsSuccessfully();
    }
}
```

Key features:
- Automatic injection via JUnit parameter resolver
- Type-safe builder pattern with validation
- Full support for transitive dependencies
- Singleton per test class (shared across test methods)
- Uses Gradle's `maven-publish` plugin under the hood

**Differences from [[issue #248](https://github.com/palantir/gradle-plugin-testing/issues/248)](https://github.com/palantir/gradle-plugin-testing/issues/248) proposal:**
- Uses `MavenArtifact` and `MavenCoordinate` for better Maven terminology
- **Does NOT support** Nebula's string graph syntax `"com.external:other:2.0.0 -> com.external:library:1.0.0"`
  - Instead uses explicit builder pattern: `.addDependency("group:artifact:version")`

==COMMIT_MSG==
Add MavenRepo test utility for publishing test dependencies

This adds a new MavenRepo utility that can be injected into JUnit tests to
publish Maven artifacts to a local test repository. This makes it easy to test
Gradle plugins that depend on external artifacts without requiring network
access or complex test setup.

Implementation uses Gradle's maven-publish plugin to create an isolated Gradle
project with subprojects for each published coordinate.

Closes https://github.com/palantir/gradle-plugin-testing/issues/248
==COMMIT_MSG==

## Possible downsides?
- **Performance**: Each `publish()` call executes a Gradle build. For tests publishing many artifacts, this could be slower than alternatives. However, this matches Nebula's implementation and ensures proper Maven repository structure.


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
```diff
@@ -77,4 +78,12 @@ default GradleFile prependLine(@Language("Gradle") @PrintFormat @FormatString St
     default GradleFile prependLine(@Language("Gradle") String line) {
         return ProjectFile.super.prependLine(line);
     }
+
+    default GradleFile withMavenRepo(MavenRepo repo) {
+        return append("""
+            repositories {
+                %s
+            }
+            """, repo.repositoryBlock());
```
felixdesouza: let's not have `MavenRepo` be responsible for rendering. Let's just grab the path of where it is and stick it in here.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/MavenRepoStore.java
```diff
@@ -0,0 +1,60 @@
...
+    private static MavenRepo initializeMavenRepo(ExtensionContext context) {
+        GradleVersion gradleVersion = GradleVersionStore.gradleVersion(context);
+        Path publisherDir = RootProjectStore.rootProject(context).path().resolve("build/mavenrepo");
+
+        clearDirectory(publisherDir);
+
+        return new MavenRepo(publisherDir, gradleVersion);
+    }
+
+    private static void clearDirectory(Path projectDir) {
```
felixdesouza: nit: we're not deleting the `projectDirectory`, we should either call it `directory` i.e. a general `clearDirectory` method, or call it what it is supposed to be `mavenRepoDirectory`.

I imagine this was lifted from `RootProjectStore`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/MavenRepoStore.java
```diff
@@ -0,0 +1,60 @@
...
+        try {
+            FileUtils.deleteDirectory(projectDir.toFile());
+            Files.createDirectories(projectDir);
+        } catch (IOException e) {
+            throw new UncheckedIOException("Failed to recreate the test directory", e);
```
felixdesouza: let's fix this in the same vein

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/MavenRepoStore.java
```diff
@@ -0,0 +1,60 @@
...
+    private MavenRepoStore() {}
+
+    static MavenRepo mavenRepo(ExtensionContext context) {
+        return context.getStore(NAMESPACE)
+                .getOrComputeIfAbsent(MAVEN_REPO_KEY, _ignored -> initializeMavenRepo(context), MavenRepo.class);
+    }
+
+    private static MavenRepo initializeMavenRepo(ExtensionContext context) {
+        GradleVersion gradleVersion = GradleVersionStore.gradleVersion(context);
+        Path publisherDir = RootProjectStore.rootProject(context).path().resolve("build/mavenrepo");
```
felixdesouza: `publisherDir` -> `repositoryDirectory`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
```diff
@@ -0,0 +1,84 @@
...
+ * void setupCommonDependencies(MavenRepo repo, RootProject root) {
+ *     repo.publish("com.palantir:service-a:1.0.0");
+ *     root.buildGradle().withMavenRepo(repo);
+ * }
+ *
+ * &#64;Test
+ * void test(MavenRepo repo) {
+ *     repo.publish(
+ *         "com.external:library:1.0.0",
+ *         "com.external:other:2.0.0 -> com.external:library:1.0.0|com.palantir:service-a:1.0.0"
```
felixdesouza: this is no longer accurate and in any case, it doesn't actually explain what it is this is doing just what bits you copy and paste.

e.g. you want to describe this along the lines of:
* a maven repository that tests can resolve against
* publishing fake modules but with the correct metadata
* to have your build under test use it, do `root.buildGradle().withMavenRepo(repo)`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
```diff
@@ -0,0 +1,84 @@
...
+        this.repoUri = repoDir.toUri();
+        this.publisher = new MavenRepoPublisher(repoDir.resolve("testgenrepo"), repoUri, gradleVersion);
+    }
+
+    /**
+     * Publishes one or more modules to the Maven repository using the builder pattern.
+     * Modules are published in the order provided to ensure dependencies are available.
+     *
+     * @param modules one or more MavenCoordinate instances
+     */
```
felixdesouza, line 67: > Publishes one or more modules to the Maven repository using the builder pattern.

this sentence doesn't make any sense, the builder pattern is not even used here ?

> Modules are published in the order provided to ensure dependencies are available.

What does the order have to do with dependencies being available?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
```diff
@@ -0,0 +1,84 @@
...
+    public MavenRepo(Path repoDir, GradleVersion gradleVersion) {
+        this.repoUri = repoDir.toUri();
+        this.publisher = new MavenRepoPublisher(repoDir.resolve("testgenrepo"), repoUri, gradleVersion);
+    }
+
+    /**
+     * Publishes one or more modules to the Maven repository using the builder pattern.
+     * Modules are published in the order provided to ensure dependencies are available.
+     *
+     * @param modules one or more MavenCoordinate instances
```
felixdesouza:
```suggestion
     * @param modules one or more {@link MavenCoordinate} instances
```

felixdesouza: I think a good rule of thumb, is to look at the rendered javadoc in your IDE. If you have that, you can see where things are missing i.e. there's a reference to something, but I can't click it because there's no link etc. There's something that should be typefaced in code because it's not a random word in the sentence. We may not use javadoc directly via the html, but it still gets rendered in the IDE

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
```diff
@@ -0,0 +1,84 @@
...
+ * void test(MavenRepo repo) {
+ *     repo.publish(
+ *         "com.external:library:1.0.0",
+ *         "com.external:other:2.0.0 -> com.external:library:1.0.0|com.palantir:service-a:1.0.0"
+ *     );
+ * }
+ * </pre>
+ */
+public final class MavenRepo {
+    private final URI repoUri;
```
felixdesouza: does this have to be a `URI`, it should just be fine with a `Path` no?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
```diff
@@ -0,0 +1,84 @@
...
+        publisher.publish(modules);
+    }
+
+    /**
+     * Returns a Gradle repository block configuration string that can be added to build.gradle in the form:
+     * <pre>{@code
+     * maven { url = uri('repoUri') }
+     * }</pre>
+     */
+    public String repositoryBlock() {
```
felixdesouza: see earlier comment, it's a little weird for this section to do this, let's just have the `GradleFile` take the path out of this class

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
```diff
@@ -0,0 +1,123 @@
...
+                    }
+                }
+            }
+            """, mavenRepoUrl, mavenRepoUrl);
+    }
+
+    void publish(List<MavenCoordinate> modules) {
+        modules.forEach(module -> {
+            createSubproject(module, subprojectName(module));
+            runPublish(module);
```
felixdesouza: instead of running `n` individual builds, why not just run one build with either `n` tasks explicitly listed, or just `publish` ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
```diff
@@ -0,0 +1,123 @@
...
+                                maven(MavenPublication) {
+                                    artifactId = '%s'
+                                    from components.java
+                                }
+                            }
+                        }
+                        """,
+                        module.group(),
+                        module.version(),
+                        module.targetCompatibility().name(),
```
felixdesouza: should this not be `sourceCompatibility` ?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/MavenRepoUsageTest.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Test
+    void dependency_validation_rejects_invalid_format() {
+        assertThatThrownBy(() -> MavenCoordinate.from("com.external:library:1.0.0")
+                        .addDependencies("com.external:lib")
+                        .build())
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("Coordinate must be in format 'group:artifact:version'")
+                .hasMessageContaining("com.external:lib");
```
felixdesouza: you can just do:

```java
assertThatIllegalArgumentException()
  .isThrownBy(() -> ...)
  .hasMessageContaining("...")
  .hasMessageContaining("...");
```

I've always found the `assertThatThrownBy` clunky and doesn't really read well.

If you have an `Exception` that's not that as ubiquitous as `IllegalArgumentException` you can also do:

```java
assertThatExceptionOfType(MySpecialException.class)
  .isThrownBy(...)
  // other assertions
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+import org.gradle.api.JavaVersion;
+import org.immutables.value.Value;
+
+/**
+ * Represents a Maven coordinate that can be published to a test repository.
+ * Use the builder pattern via {@link #from(String)} to create coordinates with dependencies.
+ */
+@Value.Immutable
+public interface MavenCoordinate {
```
felixdesouza: This is combining multiple concepts. A maven coordinate is just a group, artifact and version. It's the maven POM artifact that contains dependencies and the jar files that are the "files" in the maven repository.

I think we should split it into two:
* `MavenCoordinate` where it's just `group`, `artifact` and `version`.
* `MavenArtifact` which has a `MavenCoordinate`, and optionally has dependencies and other things we want to associate with it.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
```diff
@@ -0,0 +1,71 @@
...
+    Splitter COORDINATE_SPLITTER = Splitter.on(':').trimResults();
+
+    String group();
+
+    String artifact();
+
+    String version();
+
+    List<String> dependencies();
```
felixdesouza: these should just be `List<MavenCoordinate>` in the new world, then you don't need to have a separate `@Value.Check` for each dependency, because `MavenCoordinate` will do it upon instantiation

felixdesouza: that's not to say you can't have a builder method on the `MavenArtifact` or `MavenModule` that takes a `String`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
```diff
@@ -0,0 +1,71 @@
...
+        return JavaVersion.VERSION_1_8;
+    }
+
+    @Value.Check
+    default void validate() {
+        dependencies().forEach(MavenCoordinate::validateCoordinate);
+    }
+
+    /**
+     * Creates a coordinate builder from a coordinate string in the format "group:artifact:version".
```
felixdesouza: please type face javadoc correctly

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    static Builder from(String coordinate) {
```
felixdesouza: yeah, I was initially not a fan of this builder, because it also meant that you could change the value of either of the group, artifact or version, which is weird, and then that's when I realised these should be separate types.

you could still keep this method with the same parameter, but parse it as a `MavenCoordinate` and store it in the builder.

Also, considering that most things may not have dependencies, have a convenience method that doesn't require you to tack on `.build()` at the end of each call.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
```diff
@@ -0,0 +1,71 @@
...
+    String artifact();
+
+    String version();
+
+    List<String> dependencies();
+
+    @Value.Default
+    default JavaVersion targetCompatibility() {
+        return JavaVersion.VERSION_1_8;
+    }
```
felixdesouza: can we just not have this? do we even to need it? If we run into issues with it, we can always add it back? wdyt?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
felixdesouza: line breaks do not indicate that line breaks will be rendered, you need to add `<p>`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
```diff
@@ -0,0 +1,84 @@
...
+ * }
+ * </pre>
+ */
+public final class MavenRepo {
+    private final URI repoUri;
+    private final MavenRepoPublisher publisher;
+
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public MavenRepo(Path repoDir, GradleVersion gradleVersion) {
+        this.repoUri = repoDir.toUri();
```
felixdesouza: can we have this be a subpath ? i.e. so the tree looks like:

```
<test-root>
|- build
  |- mavenRepo
    |- repositoryRoot
    |- testgenrepo
```

It delineates what is part of the generator project and what is actually the maven repository. It's one thing I did not like from the nebula version

felixdesouza: on a similar vein, let's call `testgenrepo` something useful

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
felixdesouza: All of this becomes apparent if you try to look at the docs in your IDE

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
felixdesouza: what happens if you try to publish the same module but different versions? you would have to do them separately or? I suppose running them one by one kinda gets you the right thing, but feels a bit meh

felixdesouza: scratch that, the version is part of the name

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
```diff
@@ -0,0 +1,123 @@
...
+                        module.artifact());
+    }
+
+    private void runPublish(MavenCoordinate module) {
+        String taskPath = ":" + subprojectName(module) + ":publishMavenPublicationToMavenRepository";
+        gradleInvoker.withArgs(taskPath, "--quiet").buildsSuccessfully();
+    }
+
+    private static String subprojectName(MavenCoordinate module) {
+        return module.group() + "." + module.artifact() + "_" + module.version().replace('.', '_');
```
felixdesouza: we replace all of the dots with `_`, should we be doing the same thing here between group and artifact?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/MavenRepoUsageTest.java
```diff
@@ -0,0 +1,167 @@
...
+            dependencies {
+                implementation 'com.external:top:1.0.0'
+            }
+            """);
+
+        InvocationResult result = gradle.withArgs("dependencies").buildsSuccessfully();
+
+        assertThat(result.output()).contains("com.external:top:1.0.0");
+        assertThat(result.output()).contains("com.external:middle:1.0.0");
+        assertThat(result.output()).contains("com.external:leaf:1.0.0");
```
felixdesouza: nit: can we just chain these together?

```java
assertThat(result.output())
  .contains("com.external:top:1.0.0")
  .contains("com.external:middle:1.0.0")
  .contains("com.external:leaf:1.0.0")
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
FinlayRJW: fix in 27ebd8a041b42edd71df011533af695d8185a635

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
FinlayRJW: path works! fixed in 27ebd8a041b42edd71df011533af695d8185a635

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
FinlayRJW: renamed to `repositoryPublisherProject` and restuctured in 27ebd8a041b42edd71df011533af695d8185a635

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/MavenRepoStore.java
FinlayRJW: fixed in 84972844b822e4c6e9f9daa288fb4bcbdc0c0e2a

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/MavenRepoStore.java
FinlayRJW: yeah I just copied it - should have updated sorry! fixed in 84972844b822e4c6e9f9daa288fb4bcbdc0c0e2a

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
FinlayRJW: I ripped this wholesale from nebula but don't think we use it anywhere will remove for now and we can always add back if find a use

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
FinlayRJW: yes probably! removing targetCompatibility so wont add sourceCompatibility in

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
FinlayRJW: fixed in 09fef796258972c22785db5d04ed2e2d35a8f500

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
FinlayRJW: updated in 09fef796258972c22785db5d04ed2e2d35a8f500 think it should be better now :)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
FinlayRJW: yep your totally right have split into two and the usage feels much better see 09fef796258972c22785db5d04ed2e2d35a8f500

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
FinlayRJW: fixed this and added string overloads so can still pass in as strings

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
FinlayRJW: fixed in 09fef796258972c22785db5d04ed2e2d35a8f500

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/MavenRepoUsageTest.java
FinlayRJW: fixed in f6f9738cee653aa7489c008214f082907cc3ac26

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/MavenRepoUsageTest.java
FinlayRJW: nice! didnt know about that - fixed in f6f9738cee653aa7489c008214f082907cc3ac26

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
FinlayRJW: I think your right and we can run a single `publishMavenPublicationToMavenRepository`

FinlayRJW: fix in [8c3f798](https://github.com/palantir/gradle-plugin-testing/pull/250/commits/8c3f79811ea2d08079609eeef6abc194e42190e1)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+import java.util.List;
+import org.immutables.value.Value;
+
+/**
+ * Represents a Maven coordinate ({@code group:artifact:version}).
+ */
+@Value.Immutable
+public interface MavenCoordinate {
```
CRogers: Why not make this a `record` (with the restricted api thing)? I doesn't look like we're using builders here anyway.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+import java.util.List;
+import org.immutables.value.Value;
+
+/**
+ * Represents a Maven coordinate ({@code group:artifact:version}).
+ */
+@Value.Immutable
```
CRogers: By default `@Value.Immutable` makes the immutable class the same visibility the defining class. So public here, which is not what we wat. So every time you use it you use in a public api you need to use a `@Value.Style(visibility = ` to stop this. Probably better to just use records if we can.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenArtifact.java
```diff
@@ -0,0 +1,62 @@
...
+package com.palantir.gradle.testing.maven;
+
+import java.util.List;
+import org.immutables.value.Value;
+
+/**
+ * Represents a Maven artifact that can be published to a test repository.
+ * Contains a coordinate and optional dependencies.
+ */
+@Value.Immutable
```
CRogers, line 27: Needs `@Value.Style`. You probably want to copy what the internal `immutables-style` repo has.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/MavenRepoUsageTest.java
```diff
@@ -0,0 +1,173 @@
...
+            }
+            """);
+
+        InvocationResult result = gradle.withArgs("dependencies").buildsSuccessfully();
+
+        assertThat(result.output()).contains("com.palantir:service-a:1.0.0");
+    }
+
+    @Test
+    void simple_dependency_in_buildscript(RootProject root, MavenRepo repo, GradleInvoker gradle) {
```
CRogers: I would order it `GradleInvoker`, `RootProject`, `MavenRepo`. Try to keep it consistent, especially if AI may end generating code off of these tests/examples.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/MavenRepoUsageTest.java
```diff
@@ -0,0 +1,173 @@
...
+        root.buildGradle().prepend("""
+            buildscript {
+                // .withMavenRepo(repo) only adds to the main repo block so need to manually add
```
CRogers: I wonder if we should add a `.withMavenRepoInBuildscript` too? Tbh, I'm not sure what the use case is for being able to publish these to buildscripts are? Perhaps we can just leave out that feature/not test it here?

FinlayRJW: Removed for now if we go the https://github.com/palantir/gradle-plugin-testing/pull/255 route we will want to re-work this anyway

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenArtifact.java
FinlayRJW: added ImmutablesStyle

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
FinlayRJW: converted to a record

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepo.java
```diff
@@ -0,0 +1,83 @@
...
+ * }
+ * </pre>
+ */
+public final class MavenRepo {
+    private final Path path;
+    private final MavenRepoPublisher publisher;
+
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public MavenRepo(Path repoDir, GradleVersion gradleVersion) {
+        this.path = repoDir.resolve("repositoryRoot").toAbsolutePath();
```
crogoz: "repository" is such an overloaded term, so sometimes it is confusing 😅 , maybe we should rename it to "localMavenRepository" and the repoPublisher to be something like "localMavenRepositoryPublisherProject" ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenCoordinate.java
```diff
@@ -0,0 +1,46 @@
...
+import com.google.common.base.Preconditions;
+import com.google.common.base.Splitter;
+import com.google.errorprone.annotations.RestrictedApi;
+import com.palantir.gradle.testing.RestrictedCreation;
+import java.util.List;
+
+/**
+ * Represents a Maven coordinate ({@code group:artifact:version}).
+ */
+public record MavenCoordinate(String group, String artifact, String version) {
```
crogoz, line 28: might be worth adding something like `asString` method that can be used to get the formattedCoordinate to be used in the build.gradle eg. [here](https://github.com/palantir/gradle-plugin-testing/pull/250/files#diff-79fdc0bcf85ccfdcbfdb0fc9760c15be2b57913a8cdbe3f3f040b79d68075569R65)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/MavenRepoParameterResolver.java
```diff
@@ -0,0 +1,33 @@
...
+package com.palantir.gradle.testing.junit;
+
+import com.palantir.gradle.testing.maven.MavenRepo;
+import java.util.Optional;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ParameterContext;
+
+final class MavenRepoParameterResolver implements TerseParameterResolver {
+    @Override
+    public Optional<Object> parameter(ParameterContext parameterContext, ExtensionContext extensionContext) {
```
crogoz, line 27: I think atm I can pass multiple MavenRepo parameters to test methods/beforeEach, maybe we should throw if people try to add multiple parameters of type MavenRepo ?

docs/testing-guide.md
```diff
@@ -444,6 +447,41 @@ void access_build_output(GradleInvoker gradle, RootProject project) {
...
+        dependencies {
+            implementation 'com.example:advanced:2.0.0'
+        }
+        """);
+
+    gradle.withArgs("build").buildsSuccessfully();
+}
+```
+
+The same `MavenRepo` instance is shared across all test methods in a class. Artifacts published in `@BeforeEach` remain available for subsequent tests.
```
crogoz:
```suggestion
The `MavenRepo` instance is shared across all test methods in a class. Artifacts published in lifecycle methods (eg. `@BeforeEach`) remain available for subsequent tests.
```

docs/testing-guide.md
```diff
@@ -444,6 +447,41 @@ void access_build_output(GradleInvoker gradle, RootProject project) {
...
+            .coordinate("com.example:advanced:2.0.0")
+            .addDependency("com.example:base:1.0.0")
+            .addDependency("com.google.guava:guava:32.1.0-jre")
+            .build()
+    );
+
+    // Configure build to use the repository
+    root.buildGradle().plugins().add("java-library");
+    root.buildGradle().withMavenRepo(repo);
+    root.buildGradle().append("""
```
crogoz: I'd keep the comment closer to the mavenRepo configuration, esp. if we want to have this guide used by our automation.

```suggestion
    root.buildGradle().plugins().add("java-library");
    
    // Configure build to use the repository
    root.buildGradle().withMavenRepo(repo);
    root.buildGradle().append("""
```

crogoz (approved):

PR #251, opened 2025-10-30
Simplify project naming

FinlayRJW:
## Before this PR

Parameter names controlled project naming with confusing magic behavior:

1. **RootProject naming**: Each RootProject injection would rename the root project based on the parameter name, causing unpredictable names across different injections.

2. **Magic suffix stripping**: SubProject parameters ending in "Project" would have that suffix stripped (`SubProject assetProject` → `:asset`), while others wouldn't (`SubProject service` → `:service`).

## After this PR

1. **RootProject always named "root"**: All RootProject parameters create a project named "root" by default. Users can explicitly rename via `rootProject.settingsGradle().rootProjectName("custom")`.

2. **SubProject uses exact parameter names**: No suffix stripping. `SubProject apiService` → `:apiService`.

## Possible downsides?

**Breaking change**: Existing tests need updates:
- Tests with `SubProject fooProject` expecting `:foo` must change to `SubProject foo`
- Tests with different RootProject parameter names expecting different root project names must use explicit `rootProject.settingsGradle().rootProjectName("name")`


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
```diff
@@ -26,30 +26,35 @@ final class GradleProjectParameterResolver implements TerseParameterResolver {
...
+        rootProject.settingsGradle().rootProjectName("root");
         return rootProject;
     }
 
-    private String projectNameFromParameter(ParameterContext parameterContext) {
-        String paramName = parameterContext.getParameter().getName();
-
-        if (paramName.endsWith("Project")) {
-            return paramName.substring(0, paramName.lastIndexOf("Project"));
+    private void validateSubProjectInjectedInTestMethod(
```
FinlayRJW: I'm not 100% sure about this - it could be a bit confusing for a user to create a subproject in `@BeforeEach` and then have it be used in the test due to variable name matching - but I think its also confusing that it doesn't 😆 - maybe we do want to just be explicit and use `@Named `

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
```diff
@@ -18,38 +18,48 @@
...
 
-    private String projectNameFromParameter(ParameterContext parameterContext) {
-        String paramName = parameterContext.getParameter().getName();
-
-        if (paramName.endsWith("Project")) {
-            return paramName.substring(0, paramName.lastIndexOf("Project"));
-        }
-
-        return paramName;
+    private boolean isTestMethod(ParameterContext parameterContext) {
```
FinlayRJW: This covers all the cases I could think of

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
```diff
@@ -18,38 +18,48 @@
...
+                throw new IllegalStateException("""
+                    SubProject parameters can only be injected in test methods \
+                    (@Test, @ParameterizedTest, @RepeatedTest, @TestFactory, etc.), \
+                    not in lifecycle methods like @BeforeEach or @AfterEach. \
+                    Use RootProject.subproject("name") explicitly in lifecycle methods instead. \
+                    Found SubProject parameter '%s' in method '%s'\
+                    """.formatted(
+                                parameterContext.getParameter().getName(),
+                                parameterContext.getDeclaringExecutable().getName()));
+            }
```
CRogers: I don't really think we need this check; at least if we're getting people to understand that `SubProject foo` makes a subproject called `foo`, it's not too bad to let people use `SubProject` in `beforeEach`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
CRogers: The problem is there's always going to be more... not sure if we'd ever used it, but jqwik property based tests use `@Property`. It's going to be hard to maintain a comprehensive list, so I just wouldn't.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
```diff
@@ -18,38 +18,48 @@
...
         }
 
         return Optional.empty();
     }
 
-    private RootProject rootProjectFor(ParameterContext parameterContext, ExtensionContext extensionContext) {
+    private RootProject rootProjectFor(ExtensionContext extensionContext) {
         RootProject rootProject = RootProjectStore.rootProject(extensionContext);
-        rootProject.settingsGradle().rootProjectName(projectNameFromParameter(parameterContext));
+        rootProject.settingsGradle().rootProjectName("root");
```
CRogers: This is still problematic and we should write a test. The problem is that every time the parameter is resolved the name will be set to `root`. So imagine this situation:

```java
@BeforeEach
void beforeEach(RootProject rootProject) {
    rootProject.settingsGradle().rootProjectName("something-else");
}

@Test
void test(RootProject rootProject) {
    // now rootProject is back to `root`
}
```

We should just initialise it to `root` once in the store when the root project dir is calculated the first time.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/RootProject.java
```diff
@@ -23,9 +23,8 @@
 import java.nio.file.Path;
 
 /**
- * When injected as a parameter JUnit, the parameter name will determine the project name. If the parameter name ends
- * in "Project", the project name will be the name of the parameter with the "Project" removed. Otherwise the project
- * name will just be the parameter name.
+ * When injected as a parameter in JUnit, the root project will always be named "root" by default.
```
CRogers: You have no choice but to inject the `RootProject` as a param. I would just leave out "When injected as a parameter in JUnit, "

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/SubProject.java
```diff
@@ -21,11 +21,15 @@
...
- * in "Project", the project name will be the name of the parameter with the "Project" removed. Otherwise the project
- * name will just be the parameter name.
+ * When injected as a parameter in JUnit test methods, the parameter name will be used as the project name exactly.
+ * For example, {@code SubProject apiService} creates a subproject named "apiService".
  * <br>
- * When injected as a parameter by JUnit, the subproject will be a direct suproject of the root project.
+ * SubProject parameters can only be injected in test methods
+ * ({@code @Test}, {@code @ParameterizedTest}, {@code @RepeatedTest}, {@code @TestFactory}, etc.),
+ * not in lifecycle methods like {@code @BeforeEach} or {@code @AfterEach}.
+ * In lifecycle methods, create subprojects explicitly using {@code rootProject.subproject("name")}.
```
CRogers: This can go if we remove the check in lifecycle methods.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
FinlayRJW: good catch! should have  thought of that - fixed in [ff4ae56](https://github.com/palantir/gradle-plugin-testing/pull/251/commits/ff4ae56a2b61af843c6a0f542ef86df2735505d9)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
FinlayRJW: have removed this check as I think your right that it would be unsustainable and potentially confusing

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleProjectParameterResolver.java
FinlayRJW: removed in [1640180](https://github.com/palantir/gradle-plugin-testing/pull/251/commits/1640180b54beabdb2b1a79466e5b8be8d9225e93)

CRogers (approved):

PR #261, opened 2025-11-05
Add structured plugins block API

FinlayRJW:
## Before this PR
Plugin management in Gradle files required manual string manipulation, leading to potential errors:
- Plugins block could be placed in wrong location
- Old-style `apply plugin:` syntax could be mixed with modern `plugins {}` block
- No validation of plugin declarations
- Easy to accidentally create malformed plugins blocks

## After this PR
Structured, type-safe API for managing Gradle plugins:
- `buildGradle().plugins().add("java")` - Add plugin without version
- `buildGradle().plugins().addWithoutApply("plugin-id")` - Add plugin with `apply false`

## Important Behavioral Note

**Prepending after plugins block:** If you call `.prepend()` after using `.plugins().add()`, the prepended content will appear before the plugins block (at the very top of the file). This means the plugins block will no longer be at the top as expected. If you need content to appear after the plugins block, use `.append()` instead. We need to allow this behaviour incase users need to add a `buildscript` above the plugins block - if / when we manage `buildscript` and `pluginsManagement` positions this can be fixed

Example:
```java
buildGradle().plugins().add("java");
buildGradle().prepend("// This will be at the top\n");

// Results in:
// // This will be at the top
// plugins {
//     id 'java'
// }
```

## Possible Downsides

- Tests using `append("apply plugin: ...")` must migrate to `plugins().add(...)` although should be automatic via errorprone
- Direct manipulation of plugins block no longer allowed (by design)
- Slightly more verbose for simple cases


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,133 @@
...
+        if (buildscriptMatcher.find()) {
+            int insertPosition = buildscriptMatcher.end();
+            String suffix = text.substring(insertPosition).replaceFirst("^\n", "");
+            return text.substring(0, insertPosition) + "\n" + pluginsBlock + suffix;
+        }
+
+        return pluginsBlock + text;
+    }
+
+    static void assertNoManipulation(String oldText, String newText) {
```
FinlayRJW: This is a breaking change but think it is wise to force users / AI to use this if we want to add the smart version part in the future

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/StructuredGradleFile.java
```diff
@@ -0,0 +1,62 @@
...
+ * Gradle file with structured management of top-level blocks (plugins).
+ * Prevents direct manipulation of these blocks through {@code overwrite()} or {@code edit()}.
+ * <p>
+ * Example: {@code rootProject.buildGradle().plugins().add("java")}
+ *
+ * @see BuildGradleFile
+ * @see SettingsGradleFile
+ * @see PluginsBlock
+ */
+public interface StructuredGradleFile extends GradleFile {
```
FinlayRJW: Not sure if we are okay putting directly onto the `GradleFile` as a plain gradle file doesn't need these restrictions?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/StructuredGradleFile.java
```diff
@@ -0,0 +1,62 @@
...
+    default PluginsBlock plugins() {
+        return new PluginsBlock(editor -> {
+            GradleFile.super.overwrite(editor.edit(currentText()));
+            return this;
+        });
+    }
+
+    @Override
+    default StructuredGradleFile overwrite(String text) {
+        return edit(_ignored -> text);
```
FinlayRJW: override so everything goes via our edit function with validation

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private String insertPluginsBlock(String text, List<Plugin> plugins) {
+        String pluginsContent =
+                plugins.stream().map(plugin -> "    " + plugin.toDeclaration()).collect(Collectors.joining("\n"));
+
+        String pluginsBlock = "plugins {\n" + pluginsContent + "\n}\n";
+
+        Matcher buildscriptMatcher = BUILDSCRIPT_BLOCK_PATTERN.matcher(text);
+
+        if (buildscriptMatcher.find()) {
```
FinlayRJW, line 105: plugins block must go after buildscript and above all else

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
```diff
@@ -43,7 +43,7 @@ default T overwrite(@FormatString String text, Object... args) {
     }
 
     default T append(String text) {
-        writeString(path(), text, StandardOpenOption.APPEND);
+        edit(existingText -> existingText + text);
```
FinlayRJW: save us form also having to override append in `StructuredGradleFile`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/PluginsBlock.java
```diff
@@ -0,0 +1,131 @@
...
+     * Adds a plugin with a version. Example: {@code add("com.example.plugin", "1.0.0")}
+     */
+    public PluginsBlock add(@Language("Gradle") String pluginId, String version) {
+        return addPlugin(new Plugin(pluginId, Optional.of(version), Optional.empty()));
+    }
+
+    /**
+     * Adds a plugin with apply false. Example: {@code addWithoutApply("com.example.plugin")}
+     */
+    public PluginsBlock addWithoutApply(@Language("Gradle") String pluginId) {
```
CRogers: What use does

```
plugins {
   id 'foo' apply false
}
```

have? Tbh, idk - can we just not have this one?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/PluginsBlock.java
```diff
@@ -0,0 +1,131 @@
...
+        return addPlugin(new Plugin(pluginId, Optional.empty(), Optional.of(false)));
+    }
+
+    /**
+     * Adds a plugin with version and apply false. Example: {@code addWithoutApply("com.example.plugin", "1.0.0")}
+     */
+    public PluginsBlock addWithoutApply(@Language("Gradle") String pluginId, String version) {
```
CRogers: Tbh, I'm not sure this will be useful in tests either? Might as well not have it in the API until we need it?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/PluginsBlock.java
```diff
@@ -0,0 +1,131 @@
...
+     * Adds a plugin without a version. Example: {@code add("java")}
+     */
+    public PluginsBlock add(@Language("Gradle") String pluginId) {
+        return addPlugin(new Plugin(pluginId, Optional.empty(), Optional.empty()));
+    }
+
+    /**
+     * Adds a plugin with a version. Example: {@code add("com.example.plugin", "1.0.0")}
+     */
+    public PluginsBlock add(@Language("Gradle") String pluginId, String version) {
```
CRogers: I actually think we don't want to add this one, as it will be forever stuck at a fixed version. Ideally, any time we let people add external plugin versions they should do it in a way that enables it to be upgraded automatically. Either the `gradlePluginForTesting` approach (which I'm still not 100% sure about) or we don't let people give a version like this, instead reading the version from a file that can be upgraded by excavator.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/PluginsBlock.java
```diff
@@ -0,0 +1,131 @@
...
+import org.intellij.lang.annotations.Language;
+
+/**
+ * Structured API for managing the {@code plugins {}} block in Gradle files.
+ * <p>
+ * Example: {@code rootProject.buildGradle().plugins().add("java").add("application")}
+ *
+ * @see StructuredGradleFile#plugins()
+ */
+public record PluginsBlock(Function<FileEditor, ? extends GradleFile> editorFunction) {
```
CRogers: I would call this `Plugins` rather than `PluginBlock`. We never know, we may want to return the `apply plugin:` approach for some reason which be odd to still be called `PluginBlock` (ie the `Block` part is an implementation detail).

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/PluginsBlock.java
FinlayRJW: Got this from the shadow jar test https://github.com/palantir/gradle-shadow-jar/pull/853/files#diff-01fefc1517160ce62174a4a1480449d2c023c33abaff0027596733601785fb0aL52 - can remove if we don't think it is needed

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/StructuredGradleFile.java
CRogers: I think we should just keep this in `GradleFile` tbh. I don't think there is a reason to separate the two? The only reason I think could be the difference between project/settings `build.gradle`s that have plugins blocks vs other gradle files that are `apply from:`. I don't think separating them like this adds much value though - people can just not call `.plugins()`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/PluginsBlock.java
```diff
@@ -0,0 +1,131 @@
...
+    private static final Pattern BUILDSCRIPT_BLOCK_PATTERN =
+            Pattern.compile("buildscript\\s*\\{(?:[^{}]|\\{[^{}]*})*}", Pattern.DOTALL);
+
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public PluginsBlock {}
+
+    /**
+     * Adds a plugin without a version. Example: {@code add("java")}
+     */
+    public PluginsBlock add(@Language("Gradle") String pluginId) {
```
CRogers: I don't think this should be `@Language("Gradle")`. The plugin id is not Gradle source code. I think it may produce some sort of sensible syntax highlighting, but only by accident - the `com.blah.whatever` is valid groovy, but something like `com.blah.what-ever-with-hyphens` is not.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -139,19 +139,4 @@ void root_project_name_should_not_be_reset_to_root_in_test_method(
             gradle.withArgs().buildsSuccessfully().assertThat().output().contains("project name: something-else");
         }
     }
-
-    @Nested
-    class BeforeEachChecks {
```
FinlayRJW, line 144: removed as plugins block is tested in the plugins block tests now

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugin.java
```diff
@@ -0,0 +1,65 @@
...
+ * <p>
+ * Example: {@code new Plugin("java", Optional.of("1.0.0"), Optional.of(false))}
+ *
+ * @see PluginsBlock#add(String)
+ */
+record Plugin(String id, Optional<String> version, Optional<Boolean> apply) {
+    private static final Pattern PLUGIN_PATTERN = Pattern.compile(
+            "id\\s+'([^']+)'(?:\\s+version\\s+'([^']+)')?(?:\\s+apply\\s*\\(?\\s*(true|false)\\s*\\)?)?");
+
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
```
CRogers: No need for this if it isn't public

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,116 @@
...
+            int insertPosition = buildscriptMatcher.end();
+            String suffix = text.substring(insertPosition).replaceFirst("^\n", "");
+            return text.substring(0, insertPosition) + "\n" + pluginsBlock + suffix;
+        }
+
+        return pluginsBlock + text;
+    }
+
+    static void assertNoManipulation(String oldText, String newText) {
+        if (newText.contains("apply plugin:")) {
```
CRogers: There's also `plugins.apply(...)` and `pluginManagement.apply(...)`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
CRogers: This seems more spooky to me as it depends on `GradleFile` knowing how `ProjectFile#append` and `ProjectFile#overwrite` work. I'd just add the assertion to the `overwrite` and `append` methods' `String` arg as well and get them to call the super method after the assert.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,116 @@
...
+        if (buildscriptMatcher.find()) {
+            int insertPosition = buildscriptMatcher.end();
+            String suffix = text.substring(insertPosition).replaceFirst("^\n", "");
+            return text.substring(0, insertPosition) + "\n" + pluginsBlock + suffix;
+        }
+
+        return pluginsBlock + text;
+    }
+
+    static void assertNoManipulation(String oldText, String newText) {
```
CRogers: This needs a much better name. Manipulation of what? Perhaps `assertNoIncorrectPluginApplications`?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,116 @@
...
+        if (newText.contains("apply plugin:")) {
+            throw new IllegalStateException(
+                    "'apply plugin:' syntax is not allowed. Use gradleFile.plugins().add(\"plugin-id\")" + " instead.");
```
CRogers: One benefit to an errorprone is that the AI will be able to get an error message about, atm we don't run the tests in excavator (it is an option though) so it won't see these errors, but would if they were at compile time.

FinlayRJW: Yes I think it would probably make sense to add an errorpone and or let it run the tests once - but I think that should be a FLUP

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
FinlayRJW: how have I never come across these!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @see GradleFile#plugins()
+ */
+public record Plugins(Function<FileEditor, ? extends GradleFile> editorFunction) {
+    private static final Pattern PLUGINS_BLOCK_PATTERN = Pattern.compile("plugins\\s*\\{([^}]*)}\\n?", Pattern.DOTALL);
+    private static final Pattern BUILDSCRIPT_BLOCK_PATTERN =
+            Pattern.compile("buildscript\\s*\\{(?:[^{}]|\\{[^{}]*})*}", Pattern.DOTALL);
+
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public Plugins {}
```
CRogers: Just make this a regular class, can just make the constructor package private then

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
```diff
@@ -31,7 +47,17 @@ default GradleFile overwrite(@Language("Gradle") @PrintFormat @FormatString Stri
...
+    }
+
+    @Override
+    default GradleFile edit(FileEditor editor) {
+        String oldText = currentText();
+        @Language("Gradle")
+        String newText = editor.edit(oldText);
+        Plugins.assertNoManipulation(oldText, newText);
+        ProjectFile.super.overwrite(newText);
+        return this;
```
CRogers: It would make sense to write this in terms of `edit`, no? Like so:

```java
return ProjectFile.super.edit(text -> {
    return Plugins.assertNoIncorrectPluginApplications(editor.edit(text));
});
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/ProjectFile.java
FinlayRJW: fair I've updated so each overload checks the new text for issues - the edit compares if the old and new plugins blocks match

CRogers (dismissed):

CRogers:
:+1:

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/PluginsUsagesTest.java
```diff
@@ -0,0 +1,136 @@
...
+        rootProject.buildGradle().assertThat().hasContent("""
+            plugins {
+                id 'java'
+                id 'application'
+            }
+            """);
+    }
+
+    @Test
+    void adding_the_same_plugin_multiple_times_is_okay(RootProject rootProject) {
```
crogoz:
```suggestion
    void dedups_multiple_same_plugin_application(RootProject rootProject) {
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,107 @@
...
+        return addPlugin(new Plugin(pluginId, Optional.empty(), Optional.of(false)));
+    }
+
+    private Plugins addPlugin(Plugin plugin) {
+        file.edit(text -> addPlugin(text, plugin));
+        return this;
+    }
+
+    private String addPlugin(String originalText, Plugin plugin) {
+        List<Plugin> plugins = Stream.concat(parse(originalText).stream(), Stream.of(plugin))
```
crogoz: maybe rename this to `extractPlugins`

FinlayRJW:
Tested and autofix working on some of our migrated tests

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    private static boolean isContentMethod(MethodInvocationTree tree, VisitorState state) {
+        return Optional.ofNullable(ASTHelpers.getSymbol(tree))
+                .filter(method -> !method.getParameters().isEmpty())
+                .filter(method -> tree.getArguments().stream()
+                        .findFirst()
+                        .map(firstArg ->
+                                STRING_TYPE.matches(firstArg, state) || FILE_EDITOR_TYPE.matches(firstArg, state))
+                        .orElse(false))
```
crogoz: could we check instead if the method.enclClass() is extending ProjectFile ? I am not really sure but wouldn't something like `String something = str.method_that_returns_string("apply plugin: ....")` also match the errorProne ?

gradle-plugin-testing-error-prone/src/test/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlockTest.java
```diff
@@ -0,0 +1,528 @@
...
+                    file.append(content);
+                }
+            }
+            """);
+    }
+
+    @Test
+    void allow_all_valid_cases() {
+        test("""
+                        import com.palantir.gradle.testing.junit.GradlePluginTests;
```
crogoz: nit - the text block has too much indentation, I think you should bring it at the same level as the closing """

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugin.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+record Plugin(String id, Optional<String> version, Optional<Boolean> apply) {
+    private static final Pattern PLUGIN_PATTERN = Pattern.compile(
+            "id\\s+'([^']+)'(?:\\s+version\\s+'([^']+)')?(?:\\s+apply\\s*\\(?\\s*(true|false)\\s*\\)?)?");
+
+    /**
+     * Parses {@code "id 'java'"}, {@code "id 'plugin' version '1.0.0'"}, or {@code "id 'plugin' apply false"} into a
+     * Plugin.
+     */
+    static Plugin parse(String declaration) {
```
crogoz: this could be a constructor instead: 
```
Plugin(String declaration) {
 ...
 this(id, version, apply);
}
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,107 @@
...
+ * Example: {@code rootProject.buildGradle().plugins().add("java").add("application")}
+ *
+ * @see GradleFile#plugins()
+ */
+public final class Plugins {
+    private static final Pattern PLUGINS_BLOCK_PATTERN = Pattern.compile("plugins\\s*\\{([^}]*)}\\n?", Pattern.DOTALL);
+    private static final Pattern BUILDSCRIPT_BLOCK_PATTERN =
+            Pattern.compile("buildscript\\s*\\{(?:[^{}]|\\{[^{}]*})*}", Pattern.DOTALL);
+
+    private static final String ERROR_MESSAGE = "Plugins must be added using .plugins().add() method. "
```
crogoz: I don't think this is used

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugin.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * A single plugin declaration with ID, optional version, and optional apply setting.
+ * <p>
+ * Example: {@code new Plugin("java", Optional.of("1.0.0"), Optional.of(false))}
+ *
+ * @see Plugins#add(String)
+ */
+record Plugin(String id, Optional<String> version, Optional<Boolean> apply) {
```
crogoz: qq: should we allow hardcoding of the plugin version ? I think using the `gradlePluginTesting` workflow should be preferred.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,107 @@
...
+        file.edit(text -> addPlugin(text, plugin));
+        return this;
+    }
+
+    private String addPlugin(String originalText, Plugin plugin) {
+        List<Plugin> plugins = Stream.concat(parse(originalText).stream(), Stream.of(plugin))
+                .distinct()
+                .toList();
+
+        return insertPluginsBlock(PLUGINS_BLOCK_PATTERN.matcher(originalText).replaceFirst(""), plugins);
```
crogoz: nit - I would move this logic `PLUGINS_BLOCK_PATTERN.matcher(originalText).replaceFirst("")` to the `insertPluginsBlock` since this one is dealing with the actual logic of how to insert the plugins block.

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
FinlayRJW: yep you 100% right! I've added a check to ensure we are a ProjectFile subtype hopefully I haven't missed any edge cases!

gradle-plugin-testing-error-prone/src/test/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlockTest.java
FinlayRJW: no idea how this happened but fixed!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
FinlayRJW: whoops! good catch!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugin.java
FinlayRJW: yep good point have removed

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,314 @@
...
+            String receiver, List<PluginInfo> plugins, VisitorState state, MethodInvocationTree tree) {
+        String calls = plugins.stream()
+                .distinct()
+                .map(p -> ".%s(\"%s\")".formatted(p.apply().orElse(true) ? "add" : "addWithoutApply", p.id()))
+                .collect(Collectors.joining());
+
+        return "\n    %s.plugins()%s".formatted(receiver, calls);
+    }
+
+    private record PluginInfo(String id, Optional<String> version, Optional<Boolean> apply) {}
```
crogoz: I don't think we are using the `version` in the pluginInfo

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,314 @@
...
+                            .orElseThrow();
+                    return new PluginInfo(pluginId, Optional.empty(), Optional.empty());
+                });
+
+        Stream<PluginInfo> blocks = PLUGINS_BLOCK.matcher(cleaned).results().flatMap(block -> PLUGINS_BLOCK_ID
+                .matcher(block.group(1))
+                .results()
+                .map(i -> new PluginInfo(
+                        i.group(1),
+                        Optional.ofNullable(i.group(2)),
```
crogoz: Maybe as a FLUP, we should add a custom handler for errors of the type "plugin not found" that would give some suggestions of how to fix it (similar to: https://github.com/palantir/gradle-failure-reports/blob/7e06eccf97259143b619cdd1bccc9510da7af1ca/gradle-failure-reports/src/main/java/com/palantir/gradle/failurereports/handlers/ExecCommandNotFoundFailure.java#L6)

FinlayRJW: probably a good idea - just to check my understanding we would add a new handler to gradle-failure-reports ? if so I'll add as a FLUP

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
```diff
@@ -37,13 +41,56 @@ static boolean notWithinGradlePluginTests(Tree tree, VisitorState state) {
...
+                .isEmpty();
+    }
+
+    static Optional<ExpressionTree> findVariableInitializer(Symbol.VarSymbol var, VisitorState state) {
+        return Optional.ofNullable(ASTHelpers.findEnclosingNode(state.getPath(), MethodTree.class))
+                .flatMap(m -> m.getBody().getStatements().stream()
+                        .filter(VariableTree.class::isInstance)
+                        .map(VariableTree.class::cast)
+                        .filter(v -> var.equals(ASTHelpers.getSymbol(v)))
+                        .findFirst()
```
kelvinou01: Is `.collect(`[`MoreCollectors.toOptional`](https://guava.dev/releases/22.0/api/docs/com/google/common/collect/MoreCollectors.html#toOptional--)`)` what you want?

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,312 @@
...
+                        .map(firstArg ->
+                                STRING_TYPE.matches(firstArg, state) || FILE_EDITOR_TYPE.matches(firstArg, state))
+                        .orElse(false))
+                .filter(method -> isProjectFileSubtype(method.enclClass(), state))
+                .isPresent();
+    }
+
+    private static boolean isProjectFileSubtype(Symbol.ClassSymbol classSymbol, VisitorState state) {
+        Type classType = classSymbol.type;
+        Type projectFileType = state.getTypeFromString("com.palantir.gradle.testing.files.ProjectFile");
```
kelvinou01: nit: let's memoize this expensive call with a `Supplier`

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,312 @@
...
+            return Description.NO_MATCH;
+        }
+
+        return tree.getArguments().stream()
+                .findFirst()
+                .flatMap(arg -> checkForPlugins(arg, tree, state))
+                .orElse(Description.NO_MATCH);
+    }
+
+    private static boolean isContentMethod(MethodInvocationTree tree, VisitorState state) {
```
kelvinou01: Let's rename this method to something like "isMethodThatModifiesBuildGradle" or something?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
kelvinou01: Let's add this as a comment in the code!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/GradleFile.java
```diff
@@ -23,6 +23,14 @@
 import org.intellij.lang.annotations.PrintFormat;
 
 public interface GradleFile extends ProjectFile<GradleFile> {
```
kelvinou01, line 30: Let's add a doc here saying what a `GradleFile` is? Might not be obvious it's referring to `*.gradle` files 

(also explicating the tribal knowledge that we only care about Groovy and not Kotlin)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,105 @@
...
+ * Structured API for managing the {@code plugins {}} block in Gradle files.
+ * <p>
+ * Example: {@code rootProject.buildGradle().plugins().add("java").add("application")}
+ *
+ * @see GradleFile#plugins()
+ */
+public final class Plugins {
+    private static final Pattern PLUGINS_BLOCK_PATTERN = Pattern.compile("plugins\\s*\\{([^}]*)}\\n?", Pattern.DOTALL);
+    private static final Pattern BUILDSCRIPT_BLOCK_PATTERN =
+            Pattern.compile("buildscript\\s*\\{(?:[^{}]|\\{[^{}]*})*}", Pattern.DOTALL);
```
kelvinou01: very sad that java requires the double backslash/doesn't have the idea of a raw string. this regex is kinda incomprehensible (and rainbow brackets doesn't work inside strings!!)

kelvinou01: alas, can we make the regexes more readable? 

```java
// Matches: plugins { ... }
// Captures everything between the braces
private static final Pattern PLUGINS_BLOCK_PATTERN = Pattern.compile(
    "plugins"      // literal "plugins"
    + "\\s*"       // optional whitespace
    + "\\{"        // opening brace
    + "([^}]*)"    // capture group: any chars except }, greedily
    + "}"          // closing brace
    + "\\n?",      // optional newline after
    Pattern.DOTALL);

// Matches: buildscript { ... } with ONE level of nested braces allowed
private static final Pattern BUILDSCRIPT_BLOCK_PATTERN = Pattern.compile(
    "buildscript"           // literal "buildscript"
    + "\\s*"                // optional whitespace
    + "\\{"                 // opening brace
    + "(?:"                 // non-capturing group, repeated:
        + "[^{}]"           //   either: any char that's not a brace
        + "|"               //   OR
        + "\\{[^{}]*}"      //   a pair of braces with non-brace content inside
    + ")*"                  // repeat the group zero or more times
    + "}",                  // closing brace
    Pattern.DOTALL);
```

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/PluginsUsagesTest.java
```diff
@@ -0,0 +1,136 @@
...
+            // blah
+            """);
+
+        rootProject.buildGradle().assertThat().hasContent("""
+            // blah
+            plugins {
+                id 'java'
+            }
+            """);
+    }
```
kelvinou01, line 193: > plugins block must go after buildscript and above all else

Wonder if this is desired behavior then? Should blah come after the plugins block?

kelvinou01:
Also, as a general note, as we are implementing features, let's document as we go?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/PluginsUsagesTest.java
FinlayRJW: So unless we make `buildScript` also managed we need a way for users to be able to put something before the plugins block else: 

```java
rootProject.buildGradle().plugins().add("java");
        rootProject.buildGradle().prepend("""
            buildScript {...}
            """);
```

would give

```java
            plugins {
                id 'java'
            }
            buildScript {...}
```

and break - I think we can maybe fix this later if we make all the location controlled blocks managed but for now I figured this was okay and can fix when / if it is ever an issue

FinlayRJW:
> Also, as a general note, as we are implementing features, let's document as we go?

Probably - I've got a docs PR on the go that includes this so can get that up soon or seperate out and add the section on the plugins block as part of this PR?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
FinlayRJW: addressed in 5b9d2f183ef77a7b3664ab92995a1fde0f7f1866

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
FinlayRJW: addressed in 5b9d2f183ef77a7b3664ab92995a1fde0f7f1866

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
FinlayRJW: addressed in 5b9d2f183ef77a7b3664ab92995a1fde0f7f1866

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
FinlayRJW: addressed in 5b9d2f183ef77a7b3664ab92995a1fde0f7f1866

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
FinlayRJW: and in [8e087bb](https://github.com/palantir/gradle-plugin-testing/pull/261/commits/8e087bbccecea7208e53040ca804aeba5c0db48f)

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/PluginsUsagesTest.java
kelvinou01: As discussed in person, it's weird to have our invariant of "buildscript block > plugins block > everything else" depend on the order we call our file mutation methods. 

This is pretty critical for the usability and correctness of the plugin, so I'd say this is blocking!

kelvinou01:
> Probably - I've got a docs PR on the go that includes this so can get that up soon or seperate out and add the section on the plugins block as part of this PR?

Nice! Apart from enforcing ordering between plugins, buildscript, and the rest of the file, I think everything else here looks solid and is good to go.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/PluginsUsagesTest.java
FinlayRJW: [c2d1c6e](https://github.com/palantir/gradle-plugin-testing/pull/261/commits/c2d1c6e50c09785b3e8f789af7c37f61aadd2830) should fix this issue

FinlayRJW: I've decided to go the route of parsing the text on every call rather than storing the plugins / file in memory  - doing all the edits then writing when we execute gradle as it is 1. easier 2. does not require any state management and 3. makes it clear to a user in debug mode looking at the underlying files what is happening

kelvinou01:
+1

kelvinou01:
👍

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -0,0 +1,149 @@
...
+            "buildscript" // literal "buildscript"
+                    + "\\s*" // optional whitespace
+                    + "\\{" // opening brace
+                    + "(?:" // non-capturing group, repeated:
+                    + "[^{}]" //   either: any char that's not a brace
+                    + "|" //   OR
+                    + "\\{[^{}]*}" //   a pair of braces with non-brace content inside
+                    + ")*" // repeat the group zero or more times
+                    + "}", // closing brace
+            Pattern.DOTALL);
```
pkoenig10, line 56: This pattern isn't quite right. It doesn't correctly handle nested blocks.

For example, this correctly returns `true`:

```java
BUILDSCRIPT_BLOCK_PATTERN.matcher("buildscript{repositories{}}").find();
```

While this incorrectly returns `false`:

```java
BUILDSCRIPT_BLOCK_PATTERN.matcher("buildscript{repositories{maven{}}}").find();
```

pkoenig10: It seems like it might be simpler to just scan the string ourselves instead of trying to use a regex (regexs for matching balanced delimiters are notoriously obtuse).

FinlayRJW: thanks for catching this! fix here - https://github.com/palantir/gradle-plugin-testing/pull/288

PR #262, opened 2025-11-05
Make it possible to use external plugins in the `plugins {}` block and have them auto-update

FinlayRJW:
## Before this PR
The Nebula framework makes an `init.gradle` script which added everything from the containing project's `testRuntimeClasspath` to build script dependencies of the test.  This allowed tests to apply plugins.

The new junit-based framework does not do this - it uses Gradle's supplied TestKit framework, which has a `withPluginClasspath` method we use. This does not use an init script but instead writes out a `plugin-under-test-metadata.properties` at build time then uses it in the tests for the `plugins {}`  block classpath. It uses the `runtimeClasspath` rather than the `testRuntimeClasspath`. This causes tests which are converted from nebula to fail with errors like this:

```
* Exception is:
org.gradle.api.plugins.UnknownPluginException: Plugin [id: ‘com.palantir.jdks.latest’] was not found in any of the following sources:
- Gradle Core Plugins (plugin is not in ‘org.gradle’ namespace)
- Gradle TestKit (classpath: /home/circleci/...)
- Included Builds (No included builds contain this plugin)
- Plugin Repositories (plugin dependency must include a version number for this source)
```

...because we were including the dependencies for the gradle plugins using `testRuntimeOnly`, and test kit only looks at the `runtimeClasspath` rather than `testRuntimeClasspath`.

## After this PR
==COMMIT_MSG==
There is now a `gradlePluginsForTesting` configuration that allows external Gradle plugin dependencies to be used within Gradle plugin tests within the `plugins {}` block.
==COMMIT_MSG==

We augment the `plugin-under-test-metadata.properties` file with these extra jars to allow using external plugins in `plugins {}` blocks in tests.

They are now not included as `testRuntimeOnly`, which is good because they are not actually used in the actualy junit tests - they are used within Gradle invocations from tests.

This `gradlePluginForTesting` configuration is locked which makes it safe the augment the the `plugin-under-test-metadata.properties` file as any dependencies shared with `runtimeClasspath` will have the same versions.

## Possible downsides?
* Locking the `gradlePluginForTesting` configuration means we get constraints from gradle plugins we test against, which may be suprising. eg this repo has Jackson 2.18. We test with sls-packaging, which has Jackson 2.20. Since we test with Jackson 2.20 we use 2.20 in the lockfile.
* Had to bootstrap this feature, will remove after releasing.

### DIFF
--- a/gradle-plugin-testing/build.gradle
+++ b/gradle-plugin-testing/build.gradle
@@ -4,6 +4,15 @@ apply plugin: 'com.palantir.external-publish-jar'
 apply plugin: 'com.palantir.gradle-plugin-testing'
 apply plugin: 'com.palantir.external-publish-gradle-plugin'
 
+// used for self bootstrapping - remove on next release and bump own version
+configurations {
+    gradlePluginForTesting
+}
+
+tasks.named("pluginUnderTestMetadata", PluginUnderTestMetadata) {
+    it.getPluginClasspath().from(configurations.named("gradlePluginForTesting"))
+}
+
 dependencies {
     implementation project(':plugin-testing-core')
     implementation 'com.palantir.baseline:gradle-baseline-java'
@@ -18,7 +27,7 @@ dependencies {
     testImplementation 'org.spockframework:spock-core'
 
     testRuntimeOnly 'com.google.guava:guava'
-    testRuntimeOnly 'com.palantir.gradle.consistentversions:gradle-consistent-versions'
+    gradlePluginForTesting 'com.palantir.gradle.consistentversions:gradle-consistent-versions'
 }
 
 gradlePlugin {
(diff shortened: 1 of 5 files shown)


gradle-plugin-testing/build.gradle
```diff
@@ -4,18 +4,30 @@ apply plugin: 'com.palantir.external-publish-jar'
...
+
     testImplementation project(':gradle-plugin-testing-junit')
     testImplementation 'com.netflix.nebula:nebula-test'
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'org.codehaus.groovy:groovy'
     testImplementation 'org.spockframework:spock-core'
 
     testRuntimeOnly 'com.google.guava:guava'
+    gradlePluginForTesting 'com.palantir.gradle.consistentversions:gradle-consistent-versions'
     testRuntimeOnly 'com.palantir.gradle.consistentversions:gradle-consistent-versions'
```
FinlayRJW, line 33: Seem to still need this - I think the previous nebula tests still require it?

CRogers (approved):

PR #265, opened 2025-11-06
addition instructions for the test migration excavator

esword:
## Before this PR
Iterating on the LLM prompt for the migration is slow when we have to re-deploy the full excavator repo.  This allows for more rapid experimentation and tuning.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


FinlayRJW:
👍

PR #270, opened 2025-11-07
Ensure IntelliJ "Run test with coverage" works

CRogers:
## Before this PR
Currently, "Run with coverage" in IntelliJ does not work, because the testing framework will always fork the JVM unless debugging is enabled, and IntelliJ runs coverage without debug. It makes it look like there is no coverage:

Screenshot 2025-11-07 at 12 55 50

## After this PR
==COMMIT_MSG==
Ensure IntelliJ "Run test with coverage" works with new Java/JUnit Gradle plugin testing framework
==COMMIT_MSG==

We detect whether coverage is enabled and tell testkit to fork the JVM, like we currently do for debugging.

It now works:

Screenshot 2025-11-07 at 12 57 31


FinlayRJW:
👍

PR #283, opened 2025-11-13
Add structured plugins block API - Errorprone

FinlayRJW:
### Error Prone Check: `GradleTestPluginsBlock`

A new Error Prone check enforces usage of the structured API and prevents problematic patterns:

**What it catches:**
- `apply plugin: 'java'` - Old-style plugin application
- `plugins.apply('java')` - Direct plugin application
- `pluginManagement.apply('java')` - Plugin management application
- `plugins { id 'java' }` - Direct plugins block manipulation in strings
- Any variation of the above in string literals, text blocks, or lambda expressions

**Autofix capabilities:**
- Automatically converts old-style declarations to `.plugins().add()` calls
- Extracts plugins from strings and generates proper API calls
- Preserves other content in the string/lambda
- Handles multiple plugins in a single string

**Limitations of autofix:**
1. **No autofix for variable references** - When the plugin declaration is in a variable, only a warning is shown (no automatic fix)
2. **No autofix for `@FormatMethod` calls** - When using formatting methods (e.g., `append("plugins { id '%s' }", pluginId)`), autofix is disabled because the formatting logic would very complex
3. **Only works in `@GradlePluginTests`** - Check only applies to test classes annotated with `@GradlePluginTests`
4. **Library methods only** - Only validates calls to methods in `com.palantir.gradle.testing.*` package

**Example transformation:**
```java
// Before (triggers error)
file.append("""
    repositories { mavenCentral() }
    apply plugin: 'java'
    dependencies { }
    """);

// After (autofix)
file.append("""
    repositories { mavenCentral() }
    
    dependencies { }
    """);
file.plugins().add("java");
```

> Note: most of the changes in `GradlePluginTestHelpers` and `GradleTestStringFormatting` are simply extracting existing methods to the helper class so they can be used more widely

### DIFF
--- a/gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
+++ b/gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
@@ -29,7 +29,7 @@ import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
 import com.sun.tools.javac.code.Symbol;
 import java.util.Optional;
-import java.util.stream.StreamSupport;
+import one.util.streamex.StreamEx;
 
 public final class GradlePluginTestHelpers {
     private static final Matcher<Tree> WITHIN_GRADLE_PLUGIN_TESTS_CLASS = Matchers.enclosingNode(Matchers.allOf(
@@ -52,9 +52,8 @@ public final class GradlePluginTestHelpers {
 
     static Optional<ExpressionTree> findVariableInitializer(Symbol.VarSymbol var, VisitorState state) {
         return Optional.ofNullable(ASTHelpers.findEnclosingNode(state.getPath(), MethodTree.class))
-                .flatMap(m -> m.getBody().getStatements().stream()
-                        .filter(VariableTree.class::isInstance)
-                        .map(VariableTree.class::cast)
+                .flatMap(m -> StreamEx.of(m.getBody().getStatements())
+                        .select(VariableTree.class)
                         .filter(v -> var.equals(ASTHelpers.getSymbol(v)))
                         .collect(MoreCollectors.toOptional())
                         .map(VariableTree::getInitializer));
@@ -78,9 +77,8 @@ public final class GradlePluginTestHelpers {
         String methodName = methodSymbol.getSimpleName().toString();
         Symbol.ClassSymbol classSymbol = methodSymbol.enclClass();
 
-        return StreamSupport.stream(classSymbol.members().getSymbols().spliterator(), false)
-                .filter(Symbol.MethodSymbol.class::isInstance)
-                .map(Symbol.MethodSymbol.class::cast)
+        return StreamEx.of(classSymbol.members().getSymbols().spliterator())
+                .select(Symbol.MethodSymbol.class)
                 .filter(method -> method.getSimpleName().toString().equals(methodName))
                 .filter(method ->
                         ASTHelpers.hasAnnotation(method, "com.google.errorprone.annotations.FormatMethod", state))
--- a/gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
+++ b/gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
@@ -27,7 +27,10 @@ import com.sun.source.tree.MethodInvocationTree;
 import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.tree.VariableTree;
+import com.sun.source.util.TreePath;
+import com.sun.source.util.Trees;
 import com.sun.tools.javac.code.Symbol;
+import com.sun.tools.javac.processing.JavacProcessingEnvironment;
 import java.util.Optional;
 import one.util.streamex.StreamEx;
 
@@ -51,12 +54,14 @@ public final class GradlePluginTestHelpers {
     }
 
     static Optional<ExpressionTree> findVariableInitializer(Symbol.VarSymbol var, VisitorState state) {
-        return Optional.ofNullable(ASTHelpers.findEnclosingNode(state.getPath(), MethodTree.class))
-                .flatMap(m -> StreamEx.of(m.getBody().getStatements())
-                        .select(VariableTree.class)
-                        .filter(v -> var.equals(ASTHelpers.getSymbol(v)))
-                        .collect(MoreCollectors.toOptional())
-                        .map(VariableTree::getInitializer));
+        JavacProcessingEnvironment javacEnv = JavacProcessingEnvironment.instance(state.context);
+        return StreamEx.ofNullable(Trees.instance(javacEnv).getPath(var))
+                .filter(declPath ->
+                        declPath.getCompilationUnit() == state.getPath().getCompilationUnit())
+                .map(TreePath::getLeaf)
+                .select(VariableTree.class)
+                .map(VariableTree::getInitializer)
+                .findFirst();
     }
 
     static boolean hasFormatMethodOverload(MethodInvocationTree tree, VisitorState state) {
--- a/gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
+++ b/gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
@@ -39,16 +39,16 @@ public final class GradlePluginTestHelpers {
 
     private static final String LIBRARY_PACKAGE = "com.palantir.gradle.testing.";
 
-    static boolean notWithinGradlePluginTests(Tree tree, VisitorState state) {
-        return !WITHIN_GRADLE_PLUGIN_TESTS_CLASS.matches(tree, state);
+    static boolean isWithinGradlePluginTests(Tree tree, VisitorState state) {
+        return WITHIN_GRADLE_PLUGIN_TESTS_CLASS.matches(tree, state);
     }
 
-    static boolean notGradlePluginTestsLibraryMethod(MethodInvocationTree tree) {
+    static boolean isGradlePluginTestsLibraryMethod(MethodInvocationTree tree) {
         return Optional.ofNullable(ASTHelpers.getSymbol(tree))
                 .map(Symbol.MethodSymbol::enclClass)
                 .map(classSymbol -> classSymbol.getQualifiedName().toString())
                 .filter(className -> className.startsWith(LIBRARY_PACKAGE))
-                .isEmpty();
+                .isPresent();
     }
 
     static Optional<ExpressionTree> findVariableInitializer(Symbol.VarSymbol var, VisitorState state) {
--- a/gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
+++ b/gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
@@ -21,10 +21,14 @@ import com.google.errorprone.matchers.Matcher;
 import com.google.errorprone.matchers.Matchers;
 import com.google.errorprone.util.ASTHelpers;
 import com.sun.source.tree.ClassTree;
+import com.sun.source.tree.ExpressionTree;
 import com.sun.source.tree.MethodInvocationTree;
+import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.Tree;
+import com.sun.source.tree.VariableTree;
 import com.sun.tools.javac.code.Symbol;
 import java.util.Optional;
+import java.util.stream.StreamSupport;
 
 public final class GradlePluginTestHelpers {
     private static final Matcher<Tree> WITHIN_GRADLE_PLUGIN_TESTS_CLASS = Matchers.enclosingNode(Matchers.allOf(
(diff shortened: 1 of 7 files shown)


gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,312 @@
...
+            return Description.NO_MATCH;
+        }
+
+        return tree.getArguments().stream()
+                .findFirst()
+                .flatMap(arg -> checkForPlugins(arg, tree, state))
+                .orElse(Description.NO_MATCH);
+    }
+
+    private static boolean isContentMethod(MethodInvocationTree tree, VisitorState state) {
```
kelvinou01: Maybe a better name would be `isMethodWhichCouldAddPluginsManually`? It conveys why we care about string args or FileEditor args

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,312 @@
...
+        return Optional.ofNullable(ASTHelpers.getSymbol(tree))
+                .filter(method -> !method.getParameters().isEmpty())
+                .filter(method -> tree.getArguments().stream()
+                        .findFirst()
+                        .map(firstArg ->
+                                STRING_TYPE.matches(firstArg, state) || FILE_EDITOR_TYPE.matches(firstArg, state))
+                        .orElse(false))
+                .filter(method -> isProjectFileSubtype(method.enclClass(), state))
+                .isPresent();
```
kelvinou01: Can we simplify the logic like so? 

```
if symbol is null:
    return false
// do stuff with symbol
```

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,312 @@
...
+        return tree.getArguments().stream()
+                .findFirst()
+                .flatMap(arg -> checkForPlugins(arg, tree, state))
+                .orElse(Description.NO_MATCH);
+    }
+
+    private static boolean isContentMethod(MethodInvocationTree tree, VisitorState state) {
+        return Optional.ofNullable(ASTHelpers.getSymbol(tree))
+                .filter(method -> !method.getParameters().isEmpty())
+                .filter(method -> tree.getArguments().stream()
```
kelvinou01: since we already have the `Symbol` here, we can check the type of the first item in `method.getParameters()`, instead of going back to the tree:

`ASTHelpers.isSubtype(method.getParameters().get(0), stringType, state)`

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,312 @@
...
+                        .map(firstArg ->
+                                STRING_TYPE.matches(firstArg, state) || FILE_EDITOR_TYPE.matches(firstArg, state))
+                        .orElse(false))
+                .filter(method -> isProjectFileSubtype(method.enclClass(), state))
+                .isPresent();
+    }
+
+    private static boolean isProjectFileSubtype(Symbol.ClassSymbol classSymbol, VisitorState state) {
+        Type classType = classSymbol.type;
+        Type projectFileType = state.getTypeFromString("com.palantir.gradle.testing.files.ProjectFile");
```
kelvinou01: Let's memoize this expensive call with `Suppliers`?

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
kelvinou01: Actually I don't think `ASTHelpers.getSymbol(tree)` here is nullable so we don't have to check at all.

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
```diff
@@ -37,13 +41,56 @@ static boolean notWithinGradlePluginTests(Tree tree, VisitorState state) {
...
                 .map(classSymbol -> classSymbol.getQualifiedName().toString())
                 .filter(className -> className.startsWith(LIBRARY_PACKAGE))
+                .isEmpty();
+    }
+
+    static Optional<ExpressionTree> findVariableInitializer(Symbol.VarSymbol var, VisitorState state) {
+        return Optional.ofNullable(ASTHelpers.findEnclosingNode(state.getPath(), MethodTree.class))
+                .flatMap(m -> m.getBody().getStatements().stream()
+                        .filter(VariableTree.class::isInstance)
+                        .map(VariableTree.class::cast)
```
kelvinou01: Would [`StreamEx::mapPartial`](https://amaembo.github.io/streamex/javadoc/one/util/streamex/AbstractStreamEx.html#mapPartial(java.util.function.Function)) help?

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
```diff
@@ -37,13 +41,56 @@ static boolean notWithinGradlePluginTests(Tree tree, VisitorState state) {
...
                 .isPresent();
     }
 
+    private static boolean hasFormatOverloadInClass(Symbol.MethodSymbol methodSymbol, VisitorState state) {
+        String methodName = methodSymbol.getSimpleName().toString();
+        Symbol.ClassSymbol classSymbol = methodSymbol.enclClass();
+
+        return StreamSupport.stream(classSymbol.members().getSymbols().spliterator(), false)
+                .filter(Symbol.MethodSymbol.class::isInstance)
+                .map(Symbol.MethodSymbol.class::cast)
```
kelvinou01: Ditto the above with mapPartial

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
```diff
@@ -37,13 +42,56 @@ static boolean notWithinGradlePluginTests(Tree tree, VisitorState state) {
...
+    static boolean notGradlePluginTestsLibraryMethod(MethodInvocationTree tree) {
         return Optional.ofNullable(ASTHelpers.getSymbol(tree))
                 .map(Symbol.MethodSymbol::enclClass)
                 .map(classSymbol -> classSymbol.getQualifiedName().toString())
                 .filter(className -> className.startsWith(LIBRARY_PACKAGE))
+                .isEmpty();
+    }
+
+    static Optional<ExpressionTree> findVariableInitializer(Symbol.VarSymbol var, VisitorState state) {
+        return Optional.ofNullable(ASTHelpers.findEnclosingNode(state.getPath(), MethodTree.class))
```
kelvinou01: Do you only want to find variables initialized in this method? Would it be desireable to find e.g. fields on the test class as well? 

https://github.com/google/error-prone/blob/690c2c1038bdbeb1b08cd278e68a176c9140e92d/core/src/main/java/com/google/errorprone/bugpatterns/nullness/NullnessUtils.java#L639-L650

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
FinlayRJW: simplified to:

```java
    private static boolean isMethodWhichCouldAddPluginsManually(MethodInvocationTree tree, VisitorState state) {
        Symbol.MethodSymbol method = ASTHelpers.getSymbol(tree);
        if (method.getParameters().isEmpty()) {
            return false;
        }

        Type firstParamType = method.getParameters().get(0).type;
        if (!ASTHelpers.isSubtype(firstParamType, JAVA_LANG_STRING.get(state), state)
                && !ASTHelpers.isSubtype(firstParamType, FILE_EDITOR_TYPE.get(state), state)) {
            return false;
        }

        return Optional.ofNullable(ASTHelpers.getReceiver(tree))
                .map(receiver -> GRADLE_FILE_TYPE.matches(receiver, state))
                .orElse(false);
    }
```

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
FinlayRJW: yes I think that will work have show rework above!

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
FinlayRJW: I think StreamEx::select is what I want

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
FinlayRJW: seems sensible updated!

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
```diff
@@ -37,13 +43,56 @@ static boolean notWithinGradlePluginTests(Tree tree, VisitorState state) {
         return !WITHIN_GRADLE_PLUGIN_TESTS_CLASS.matches(tree, state);
     }
 
-    static boolean isGradlePluginTestsLibraryMethod(MethodInvocationTree tree, VisitorState state) {
+    static boolean notGradlePluginTestsLibraryMethod(MethodInvocationTree tree) {
         return Optional.ofNullable(ASTHelpers.getSymbol(tree))
                 .map(Symbol.MethodSymbol::enclClass)
                 .map(classSymbol -> classSymbol.getQualifiedName().toString())
                 .filter(className -> className.startsWith(LIBRARY_PACKAGE))
+                .isEmpty();
```
esword: it's useful to not in the PR when a bunch of methods have just been moved from one file to another, or, as is the case for findVariableInitializer I think, extracted from another method in another file

FinlayRJW: will update PR decription

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,377 @@
...
+            VisitorState.memoize(state -> state.getTypeFromString("java.lang.String"));
+    private static final Supplier<Type> FILE_EDITOR_TYPE = VisitorState.memoize(
+            state -> state.getTypeFromString("com.palantir.gradle.testing.files.ProjectFile$FileEditor"));
+
+    // Matches three forms of plugin declarations:
+    // 1. apply plugin: "plugin-id"
+    // 2. plugins.apply("plugin-id")
+    // 3. pluginManagement.apply("plugin-id")
+    // Captures the plugin ID in one of three groups depending on which form matched
+    private static final Pattern SIMPLE_PLUGIN_DECLARATIONS = Pattern.compile(
```
esword, line 66: only an AI comments this much. ;-)

FinlayRJW: haha - was inspired by https://github.com/palantir/gradle-plugin-testing/pull/261#discussion_r2523666082

gradle-plugin-testing-error-prone/src/test/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlockTest.java
```diff
@@ -0,0 +1,528 @@
...
+            }
+            """);
+    }
+
+    @Test
+    void allow_outside_gradle_plugin_tests() {
+        test("""
+            import com.palantir.gradle.testing.files.gradle.GradleFile;
+
+            class TestClass {
```
esword, line 134: why allow this case?

FinlayRJW: `allow_outside_gradle_plugin_tests` not in @GradlePluginTests i.e. so other random test

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
```diff
@@ -0,0 +1,377 @@
...
+            return Description.NO_MATCH;
+        }
+
+        return tree.getArguments().stream()
+                .findFirst()
+                .flatMap(arg -> checkForPlugins(arg, tree, state))
+                .orElse(Description.NO_MATCH);
+    }
+
+    private static boolean isMethodWhichCouldAddPluginsManually(MethodInvocationTree tree, VisitorState state) {
```
esword, line 151: just for readability, useful to put a method comment here for what it's expecting to catch

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradlePluginTestHelpers.java
```diff
@@ -37,13 +43,56 @@ static boolean notWithinGradlePluginTests(Tree tree, VisitorState state) {
         return !WITHIN_GRADLE_PLUGIN_TESTS_CLASS.matches(tree, state);
     }
 
-    static boolean isGradlePluginTestsLibraryMethod(MethodInvocationTree tree, VisitorState state) {
+    static boolean notGradlePluginTestsLibraryMethod(MethodInvocationTree tree) {
```
esword: naming methods with a negative is strange.  There are times when it's useful so you don't have to put ! in front of a hundred things, but I think this and the other "not" one would be better as the original "is"

FinlayRJW: Yeah I had as is but intellij suggested switching - will switch back

gradle-plugin-testing-error-prone/src/main/java/com/palantir/gradle/testing/errorprone/GradleTestPluginsBlock.java
FinlayRJW: added:

```
We want to try and catch all methods on GradleFile that might edit the file and thus might change the plugins block. Rather than just looking at a strict list we look at any method that takes a String or FileEditor as the first parameter and is called on a GradleFile instance.
```

esword (approved):

PR #287, opened 2025-11-14
Create the gradle/gradle-test-versions.yml file

kelvinou01:
## Before this PR

No way to excavate new Gradle versions for testing easily

## After this PR

Introduce a new file `gradle/gradle-test-versions.yml`:

```yaml
major-versions:
  8: 8.14.3
  9: 9.2.0-rc1
extra-versions:
- 8.8
- 8.14.2
```

Add a section to the `README` detailing how gradleVersions is set. Refer to the README for details on how the config file is used. 

==COMMIT_MSG==
Create the gradle/gradle-test-versions.yml file
==COMMIT_MSG==


gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,98 @@
...
+
+@Value.Immutable
+@JsonDeserialize(as = ImmutableGradleTestVersionsConfig.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public interface GradleTestVersionsConfig {
+    @JsonProperty("major-versions")
+    @Value.Default
+    default SortedMap<Integer, String> majorVersions() {
+        return new TreeMap<>();
+    }
```
felixdesouza: do not need to provide implementations here, just annotate it with `@Value.NaturalOrder`

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,98 @@
...
+
+    default GradleTestVersionsConfig withMajorVersion(String newMajorVersion) {
+        int majorVersion = getMajorVersion(newMajorVersion);
+
+        SortedMap<Integer, String> newMajorVersions = new TreeMap<>();
+        newMajorVersions.putAll(this.majorVersions());
+        newMajorVersions.put(majorVersion, newMajorVersion);
+
+        return builder().from(this).majorVersions(newMajorVersions).build();
```
felixdesouza: if you use the `@Value.NaturalOrder` you do not need to do all of this

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingExtension.java
```diff
@@ -35,8 +53,25 @@ public abstract class PluginTestingExtension {
 
     public PluginTestingExtension() {
         getIgnoreGradleDeprecations().convention(true);
-        // TODO(#XXX): Should this be the latest gradle 8, or maybe whatever this plugin is compiled against?
-        // or is this the set of "milestone" versions and we dynamically add the version of the consuming project?
-        getGradleVersions().convention(GradleTestVersions.DEFAULT_TEST_GRADLE_VERSIONS);
+
+        // If the config file doesn't exist, the user has to set `gradleVersions {}` in the build script
```
felixdesouza: no one is going to read this, and this convention is weird, you shouldn't need to do this surely

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingExtension.java
```diff
@@ -35,8 +53,25 @@ public abstract class PluginTestingExtension {
...
+                        "No gradleVersions were set for the PluginTestingExtension. Either set the default versions via"
+                                + " gradle/gradle-test-versions.yml, or add overrides via the gradleVersions property",
+                        e);
+            }
+            Stream<String> majorVersions =
+                    EntryStream.of(config.majorVersions()).values();
+            Stream<String> extraVersions = config.extraVersions().stream();
+            return Stream.concat(majorVersions, extraVersions).collect(Collectors.toSet());
+        });
+        getGradleVersions().convention(defaultGradleVersions);
```
felixdesouza: this is modifiable, and for existing places, will completely override what you get from this file which is pointless

What you want to do is have a separate *provider* that you wire into tests, and then eventually you phase out the `gradleVersions` property

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,289 @@
...
+import java.util.TreeSet;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+public class GradleTestVersionsConfigTest {
+
+    @Nested
+    class Deserialize {
+        private static final String NORMAL_YAML = """
```
felixdesouza: you only ever use these in a single test, please put them where they are used, it makes the test easier to read in isolation

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,289 @@
...
+            assertThat(config.majorVersions()).containsEntry(8, "8.14.2").containsEntry(9, "9.2.0");
+            assertThat(config.extraVersions()).containsExactly("8.8");
+        }
+
+        @Test
+        public void deserialize_without_extra_versions_ok() throws IOException {
+            Path yamlFile = tempDir.resolve("versions.yaml");
+            Files.writeString(yamlFile, NO_EXTRA_VERSIONS_YAML);
+
+            GradleTestVersionsConfig config = GradleTestVersionsConfig.deserialize(yamlFile);
```
felixdesouza: I think it's weird you need to write to a file to test deserialisation + part of the reason why I don't think the config class having deserialisation methods makes sense as you have to stuff like this

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,289 @@
...
+
+        @Test
+        public void deserialize_with_rc_ok() throws IOException {
+            Path yamlFile = tempDir.resolve("versions.yaml");
+            Files.writeString(yamlFile, RC_YAML);
+
+            GradleTestVersionsConfig config = GradleTestVersionsConfig.deserialize(yamlFile);
+
+            assertThat(config.majorVersions()).containsEntry(8, "8.14.2").containsEntry(9, "9.2.0-rc1");
+            assertThat(config.extraVersions()).containsExactly("8.8");
```
felixdesouza: let's not do this, we can just make the expected object and compare the two i.e.

```java
GradleTestVersionsConfig expected = GradleTestVersionsConfig.builder()
  .majorVersion(8, "8.14.2")
  .majorVersion(9, "9.2.0-rc1")
  .extraVersions("8.8")
  .build();

assertThat(config).isEqualTo(expected);
```

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,289 @@
...
+                    .majorVersions(majorVersions)
+                    .extraVersions(extraVersions)
+                    .build();
+
+            assertThat(config.majorVersions()).containsEntry(8, "8.14.2").containsEntry(9, "9.2.0");
+            assertThat(config.extraVersions()).containsExactly("8.8");
+        }
+
+        @Test
+        public void testEqualsAndHashCode() {
```
felixdesouza: you don't need to do this, immutables generates this for you and your previous tests will effectively check this for you because you're calling equals all the time

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,289 @@
...
+
+            assertThat(updated.majorVersions()).containsEntry(8, "8.14.2");
+            assertThat(updated.majorVersions()).hasSize(1);
+        }
+    }
+
+    @Nested
+    class Miscellaneous {
+        @Test
+        public void testBuilderPattern() {
```
felixdesouza: this is not a useful test

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,289 @@
...
+
+            GradleTestVersionsConfig config = GradleTestVersionsConfig.builder()
+                    .majorVersions(majorVersions)
+                    .extraVersions(new TreeSet<>())
+                    .build();
+
+            GradleTestVersionsConfig updated = config.withoutMajorVersion(9);
+
+            assertThat(updated.majorVersions()).containsEntry(8, "8.14.2");
+            assertThat(updated.majorVersions()).hasSize(1);
```
felixdesouza: once you've converted to an expected object assertion, add a description to the assert, I'm not sure what I'm supposed to be looking at here

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
felixdesouza: this is the same throughout all of these tests

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,98 @@
...
+        return new TreeSet<>();
+    }
+
+    final class Builder extends ImmutableGradleTestVersionsConfig.Builder {}
+
+    static Builder builder() {
+        return new Builder();
+    }
+
+    static GradleTestVersionsConfig deserialize(Path pathToResource) throws IOException {
```
felixdesouza: as stated elsewhere, this method is needlessly restrictive and it should be up to the consumer to deserialise the object. You shouldn't need to deal with empty handling here either.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingExtension.java
```diff
@@ -35,8 +53,25 @@ public abstract class PluginTestingExtension {
...
+                throw new UncheckedIOException(
+                        "No gradleVersions were set for the PluginTestingExtension. Either set the default versions via"
+                                + " gradle/gradle-test-versions.yml, or add overrides via the gradleVersions property",
+                        e);
+            }
+            Stream<String> majorVersions =
+                    EntryStream.of(config.majorVersions()).values();
+            Stream<String> extraVersions = config.extraVersions().stream();
+            return Stream.concat(majorVersions, extraVersions).collect(Collectors.toSet());
+        });
```
felixdesouza: `default` is not the right terminology, they're not expected to be overridden.

separately, let's make use configuration cache apis:

```java
RegularFile configFile = getBuildLayout().getRootDirectory().file("gradle/gradle-test-versions.yml");
getProviderFactory().fileContents(configFile)
  .getAsText()
  .map(PluginTestingExtension::deserialise)
  .map(GradleTestVersionsConfig::allVersions)
```

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingExtension.java
felixdesouza: this provider will merge `getGradleVersions`  -  the current modifiable property - with the new thing. You can then phase out the `getGradleVersions` by deleting it from buildscripts.

felixdesouza: if what you pass to tests is empty, a reasonable default is to use the current version of gradle

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.TreeMap;
+import java.util.TreeSet;
+import one.util.streamex.EntryStream;
+import one.util.streamex.StreamEx;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonDeserialize(as = ImmutableGradleTestVersionsConfig.class)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public interface GradleTestVersionsConfig {
```
felixdesouza, line 35: let's have an `allVersions` method

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingExtension.java
```diff
@@ -16,11 +16,29 @@
...
+import one.util.streamex.EntryStream;
 import org.gradle.api.provider.Property;
+import org.gradle.api.provider.Provider;
+import org.gradle.api.provider.ProviderFactory;
 import org.gradle.api.provider.SetProperty;
+import org.gradle.initialization.layout.BuildLayout;
 
 public abstract class PluginTestingExtension {
     public static final String EXTENSION_NAME = "gradleTestUtils";
+    public static final Path CONFIG_FILE = Path.of("gradle", "gradle-test-versions.yml");
```
felixdesouza: private

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -129,9 +128,27 @@ class PluginTestingPluginIntegrationSpec extends AbstractTestingPluginSpec {
         version << GradleTestVersions.gradleVersionsForTests
     }
 
+    def 'fails with helpful error message if gradleVersions not set: #version'() {
```
felixdesouza: test is obsolete with a sensible default of the current gradle version

gradle-plugin-testing/build.gradle
```diff
@@ -5,17 +5,27 @@ apply plugin: 'com.palantir.gradle-plugin-testing'
...
     implementation project(':plugin-testing-core')
     implementation 'com.palantir.baseline:gradle-baseline-java'
     implementation 'com.palantir.suppressible-error-prone:gradle-suppressible-error-prone'
+    implementation 'one.util:streamex'
+
+    annotationProcessor 'org.immutables:value'
 
     compileOnly 'com.palantir.gradle.consistentversions:gradle-consistent-versions'
+    compileOnlyApi 'org.immutables:value::annotations'
+    compileOnlyApi 'org.immutables:serial'
```
felixdesouza: you don't need this dependency I don't think, you're not using any java serialisation

gradle-plugin-testing/build.gradle
```diff
@@ -5,17 +5,27 @@ apply plugin: 'com.palantir.gradle-plugin-testing'
...
+
 
     testImplementation project(':gradle-plugin-testing-junit')
     testImplementation 'com.netflix.nebula:nebula-test'
     testImplementation 'org.junit.jupiter:junit-jupiter'
     testImplementation 'org.codehaus.groovy:groovy'
     testImplementation 'org.spockframework:spock-core'
+    testImplementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml'
```
felixdesouza: you shouldn't need this surely? you've got it in your api, or is this one of the verification methods in baseline about explicit deps

README.md
```diff
@@ -125,10 +125,42 @@ class HelloWorldSpec extends IntegrationSpec {
...
 ```
-The plugin sets default versions to test against, but these can be overridden using the `gradleTestUtils` extension.  e.g.
+
+# Setting Gradle test versions
+
+There are two ways to set which Gradle versions to test against:
+
+### The `gradle/gradle-test-versions.yml` file
+
+By default, all `@GradlePluginsTests` run against the versions specified in `gradle/gradle-test-versions.yml`. Using a yaml file allows excavator to easily roll out minor version bumps, and remove deprecated major versions.
```
felixdesouza: don't think it's necessary to talk about excavator here,

felixdesouza: actually, this is fine I guess.

I think we don't want to harp on about excavator explicitly, but largely just be like: the `major-versions` block is expected to be edited by robots where it will bump the versions to test against. The `extra-versions` block is expected to be edited by humans where they need to run a particular version to test specific behaviour

README.md
```diff
@@ -125,10 +125,42 @@ class HelloWorldSpec extends IntegrationSpec {
...
+
+To test that your plugin works against a specific Gradle version e.g. 8.8, add it to `extra-versions`. 
+
+```yaml
+major-versions:
+- 8: 8.14.3
+- 9: 9.2.0-rc1
+extra-versions:
+- 8.8
+- 8.14.2
```
felixdesouza, line 129:
```suggestion
major-versions:
 8: 8.14.3
 9: 9.2.0-rc1
extra-versions:
- 8.8
- 8.14.2
```

felixdesouza: in the before,  you have a list of map, but you want a map

README.md
```diff
@@ -125,10 +125,42 @@ class HelloWorldSpec extends IntegrationSpec {
...
+> If you need to test against ext
+
 
 ```groovy
 gradleTestUtils {
     gradleVersions = ['7.6.4', '8.8']
 }
 ```
+
+Alternatively, you can set the `gradleVersions` property, which overrides the config file.
```
felixdesouza: this behaviour is not desirable! you can't ever trust that what's in your config file will ever get run. If it merges, you know what's in the config file will get run + potentially more things

README.md
```diff
@@ -125,10 +125,42 @@ class HelloWorldSpec extends IntegrationSpec {
...
+major-versions:
+- 8: 8.14.3
+- 9: 9.2.0-rc1
+extra-versions:
+- 8.8
+- 8.14.2
+```
+
+
+### Setting `gradleVersions` in the `gradleTestUtils` extension
```
felixdesouza: I think we should just say something along the lines of:


```
### [Deprecated] Setting gradle versions to test in the `gradleTestUtils` extension

You can set Gradle versions to test by setting `gradleVersions` property on the `gradleTestUtils` extension.

// example

This will merge with the what's set in the config file.

This is deprecated as the config file is the desired source of truth here.
```

README.md
```diff
@@ -125,10 +125,42 @@ class HelloWorldSpec extends IntegrationSpec {
     }
 }
 ```
-The plugin sets default versions to test against, but these can be overridden using the `gradleTestUtils` extension.  e.g.
+
+# Setting Gradle test versions
```
felixdesouza: This header doesn't make sense here I don't think. There's already a `Resolution of Gradle versions to test against` header which is describing the other half of this logic.  We should make it a h2 I think

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.SortedMap;
+import java.util.SortedSet;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import one.util.streamex.EntryStream;
+import one.util.streamex.StreamEx;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonDeserialize(as = ImmutableGradleTestVersionsConfig.class)
```
felixdesouza, line 33: you should make this serialisable as well, as excavator will write this out

felixdesouza: I am curious about how excavator will consume this, perhaps it will be fine, but there'd need to be some dependency, on `gradle-plugin-testing`

felixdesouza: that is `@JsonSerialize(as = ImmutableGradleTestVersionsConfig.class)`

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
kelvinou01: Hmm how do we get a sensible default? 
- Hardcoded "8.14.3"
- Fetch with a curl from Gradle's github releases

Both options don't seem good to me. Also, now we're unioning the config file and the extension, so I think not having a default is fine here?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
kelvinou01: yep!

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+
+    final class Builder extends ImmutableGradleTestVersionsConfig.Builder {
+        public Builder majorVersion(int majorVersion, String version) {
+            putMajorVersions(majorVersion, version);
+            return this;
+        }
+
+        public Builder extraVersion(String version) {
+            addExtraVersions(version);
+            return this;
```
felixdesouza: why do you need to add this?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+                .from(this)
+                .majorVersions(newMajorVersions)
+                .extraVersions(newExtraVersions)
+                .build();
+    }
+
+    default SortedSet<String> allVersions() {
+        Stream<String> majorVersions = EntryStream.of(this.majorVersions()).values();
+        Stream<String> extraVersions = this.extraVersions().stream();
+        return Stream.concat(majorVersions, extraVersions).collect(Collectors.toCollection(TreeSet::new));
```
felixdesouza: you can literally just do:

```java
return ImmutableSortedSet.<String>builder()
  .addAll(majorVersions().values())
  .addAll(extraVersions())
  .build()
```

It's also an immutable copy

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+    GradleTestVersionsConfig withExtraVersions(String... elements);
+
+    default GradleTestVersionsConfig withoutMajorVersion(int majorVersion) {
+        SortedMap<Integer, String> newMajorVersions = EntryStream.of(majorVersions())
+                .filterKeys(maj -> maj != majorVersion)
+                .toSortedMap();
+        SortedSet<String> newExtraVersions = StreamEx.of(extraVersions())
+                .filter(version -> {
+                    int versionMajor = getMajorVersion(version);
+                    return versionMajor != majorVersion;
```
felixdesouza: let's just inline this

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingExtension.java
```diff
@@ -41,8 +52,9 @@ public abstract class PluginTestingExtension {
     public PluginTestingExtension() {
         getIgnoreGradleDeprecations().convention(true);
         getConfigurationCacheEnabled().convention(false);
-        // TODO(#XXX): Should this be the latest gradle 8, or maybe whatever this plugin is compiled against?
-        // or is this the set of "milestone" versions and we dynamically add the version of the consuming project?
-        getGradleVersions().convention(GradleTestVersions.DEFAULT_TEST_GRADLE_VERSIONS);
+
+        Provider<Set<String>> testDriverGradleVersion = getProviderFactory()
+                .provider(() -> Set.of(GradleVersion.current().getVersion()));
```
felixdesouza: this does not need to be wrapped in a provider, just pass it in directly

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -50,9 +59,16 @@ public class PluginTestingPlugin implements Plugin<Project> {
             List.of("plugin-testing-core", "configuration-cache-spec", "gradle-plugin-testing-junit");
 
     public static final Set<String> PATCHABLE_CHECKS = Set.of("GradleTestStringFormatting", "GradleTestPluginsBlock");
+    public static final ObjectMapper YAML_MAPPER = new ObjectMapper(new YAMLFactory());
```
felixdesouza, line 63: does this need to be public?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -207,4 +224,39 @@ private static void addGradlePluginForTestingConfigurations(Project project) {
...
+    private Provider<Set<String>> combineGradleTestVersions(
+            Provider<Set<String>> versionsFromConfigFile, Provider<Set<String>> versionsFromExtension) {
+
+        Provider<Set<String>> configOrEmpty = versionsFromConfigFile.orElse(Set.of());
+        Provider<Set<String>> extensionOrEmpty = versionsFromExtension.orElse(Set.of());
+
+        return configOrEmpty.zip(extensionOrEmpty, (config, extension) -> {
+            Set<String> combined = StreamEx.of(config).append(extension).toSet();
+            if (combined.isEmpty()) {
+                throw new IllegalStateException(
```
felixdesouza: let's shift the emptiness checking into the place its used and not do it here. 

i.e. inside the junit extension that consumes this, default to the current gradle version

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -207,4 +224,39 @@ private static void addGradlePluginForTestingConfigurations(Project project) {
                     .test(scope -> scope.from(gradlePluginForTestingResolvable.getName()));
         });
     }
+
+    private Provider<Set<String>> combineGradleTestVersions(
```
felixdesouza: You're going to call `.get` on the next line where this is consumed, let's get rid of the providers and realise it here. Whenever gradle 10 comes out, we'd have to move all of this around anyway.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -207,4 +224,39 @@ private static void addGradlePluginForTestingConfigurations(Project project) {
...
+                getProviderFactory().fileContents(testVersionsConfigPath).getAsText();
+
+        Provider<GradleTestVersionsConfig> config = yamlString.map(text -> {
+            try {
+                return YAML_MAPPER.readValue(text, new TypeReference<>() {});
+            } catch (JsonProcessingException e) {
+                throw new IllegalArgumentException(e);
+            }
+        });
+        return config.map(GradleTestVersionsConfig::allVersions);
```
felixdesouza: nit: just chain these all together?

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/ConfigurationCacheTest.groovy
```diff
@@ -31,11 +31,18 @@ class ConfigurationCacheTest extends ConfigurationCacheSpec {
                 mavenCentral()
                 mavenLocal()
             }
+
+            gradleTestUtils {
+                gradleVersions = ["${version}"]
+            }
```
felixdesouza: Looking at this test, it makes no sense. There aren't any tests for the build under test to run.

Your changes also don't do anything:
* For every version in `version`, it doesn't run the build under test with the parameterised gradle version, it runs the build under test with the default version of gradle. Then any tests that are run *within* the build under test, if they use the `@GradlePluginTesting` annotation, then those tests will run with different gradle versions. But there are no tests to run.

I would be surprised if this test actually did anything i.e. if the task was just `NO-SOURCES`.

I would probably delete the test, or set it up properly, and in any case, I don't think the change you're making here is necessary

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -302,10 +339,14 @@ class PluginTestingPluginIntegrationSpec extends AbstractTestingPluginSpec {
         version << GradleTestVersions.gradleVersionsForTests
     }
 
-    void applyTestUtilsPlugin() {
+    void applyTestUtilsPlugin(String version) {
```
felixdesouza: I don't think you need to add this. You're running with the current version of Gradle. and you're not doing any multi version testing. It was only added so things wouldn't break

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.io.IOException;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+public class GradleTestVersionsConfigTest {
+    @Nested
+    class Deserialize {
+        private static GradleTestVersionsConfig deserialize(String content) throws IOException {
```
felixdesouza: private static methods at the bottom, out of the way

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,178 @@
...
+            GradleTestVersionsConfig expected = GradleTestVersionsConfig.builder()
+                    .majorVersion(9, "9.2.0")
+                    .extraVersion("9.0")
+                    .build();
+
+            assertThat(updated).isEqualTo(expected);
+        }
+
+        @Test
+        public void keeps_unrelated_extra_versions() {
```
felixdesouza: this is the same as the above test?

plugin-testing-core/build.gradle
```diff
@@ -3,6 +3,8 @@ apply plugin: 'com.palantir.external-publish-jar'
 
 dependencies {
     implementation 'com.google.guava:guava'
+    implementation 'com.fasterxml.jackson.core:jackson-databind'
+    implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml'
```
felixdesouza: why do you need these? you haven't changed anything in this project

gradle-plugin-testing/build.gradle
```diff
@@ -5,11 +5,19 @@ apply plugin: 'com.palantir.gradle-plugin-testing'
 apply plugin: 'com.palantir.external-publish-gradle-plugin'
 
 dependencies {
+    api 'com.fasterxml.jackson.core:jackson-databind'
+    api 'com.fasterxml.jackson.dataformat:jackson-dataformat-yaml'
```
felixdesouza: I think you only need the annotations to be api, and everything else can be implementation

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -129,9 +128,43 @@ class PluginTestingPluginIntegrationSpec extends AbstractTestingPluginSpec {
         version << GradleTestVersions.gradleVersionsForTests
     }
 
+    def 'use the test driver Gradle version when a version is not set: #version'() {
```
felixdesouza, line 131: this is a much better example of what the `ConfigurationCacheTest` could be doing

felixdesouza, line 131: I think you also need a test that the file does what it says i.e. can you run the build under test on a single gradle version, but set it up that the config file runs two versions. That a test you've written runs against two versions.

There's nothing here in this PR that does that. You've got deserialisation logic testing and that's it, there's nothing that indicates that you've wired things up correctly.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    static Builder builder() {
+        return new Builder();
+    }
+
+    GradleTestVersionsConfig withMajorVersions(Map<Integer, ? extends String> entries);
+
+    GradleTestVersionsConfig withExtraVersions(String... elements);
```
felixdesouza: these are not used, and in any case they're generated by immutables

README.md
felixdesouza: let's restructure this sentence. It implies that it will be up-to-date by virtue of just using the class, but negates to mention that it's actually about the wiring from `gradleVersion` and also the config file.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfig.java
kelvinou01: Wow this is really useful, thank you!

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
kelvinou01: yep, so other classes can use the same object mapper

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/ConfigurationCacheTest.groovy
kelvinou01: Hmm, the test does check the plugin logic run during configuration time for CC violations. 

But no tasks are run, so the logic in `test.doFirst` isn't tested. I'll fix this in a FLUP PR?

felixdesouza: sounds good

README.md
kelvinou01: Yep!

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
kelvinou01: Good idea, I've aded this test in PluginTestingPluginIntegrationSpec:

```groovy
    def 'use Gradle versions from both config file and extension'() {
        given: 'gradle/gradle-test-versions.yml exists, and sets both major and extra versions'
        ...

        and: 'gradleVersions is set in build script'
        ...

        when: 'running GradlePluginTests'
        ...

        then: 'test runs with versions from both config file and extension'
        result.success
        result.standardOutput.contains("Running with Gradle version: 8.14.2")
        result.standardOutput.contains("Running with Gradle version: 8.8")
        result.standardOutput.contains("Running with Gradle version: 7.6")
    }
```

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingExtension.java
```diff
@@ -30,7 +32,10 @@ public abstract class PluginTestingExtension {
 
     /**
      * Gradle versions to test against.
+     * @deprecated Use gradle/gradle-test-versions.yml to set Gradle versions to test against. This property will be
+     * removed as part of rolling out gradle/gradle-test-versions.yml.
```
felixdesouza:
```suggestion
     * @deprecated Use {@code gradle/gradle-test-versions.yml} to set Gradle versions to test against. This property will be
     * removed as part of rolling out {@code gradle/gradle-test-versions.yml}.
```

gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy
```diff
@@ -169,9 +256,13 @@ class PluginTestingPluginIntegrationSpec extends AbstractTestingPluginSpec {
...
         given:
-        prependToBuildFile('''
+        prependToBuildFile("""
             apply plugin: 'com.palantir.gradle-plugin-testing'
-        ''')
+            
+            gradleTestUtils {
+                gradleVersions = ["${version}"]
+            } 
+        """)
```
felixdesouza: undo this change

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -0,0 +1,155 @@
...
+            return PluginTestingPlugin.YAML_MAPPER.readValue(content, new TypeReference<>() {});
+        }
+    }
+
+    @Nested
+    class WithoutMajorVersion {
+        @Test
+        public void removes_major_version() {
+            GradleTestVersionsConfig config = GradleTestVersionsConfig.builder()
+                    .majorVersions(Map.of(7, "7.6.4", 8, "8.14.2", 9, "9.2.0"))
```
felixdesouza: can we just call `majorVersions` multiple times or whatever the additive version is, this just makes it hard to read

README.md
```diff
@@ -106,7 +106,52 @@ dependencies {
...
+```groovy
+gradleTestUtils {
+    gradleVersions = ['7.6.4', '8.8']
+}
+```
+
+This will merge with the what's set in the config file.
+
+> [!INFO]
+> This is deprecated as the config file is the desired source of truth here.
```
felixdesouza:
````suggestion
There are two ways to set which Gradle versions to test against:
 * setting the versions in the `gradle/gradle-test-versions.yml` file
 * setting `gradleVersions` in the the `gradleTestUtils` extension (this is deprecated)

When using `@GradlePluginTests`, the versions of Gradle used will be the union of the versions above.

### The `gradle/gradle-test-versions.yml` file

An example file:
```yaml
major-versions:
  8: 8.14.3
  9: 9.2.0-rc1
extra-versions:
- 8.8
- 8.14.2
```

The motivation for having a separate config file is to more easily allow an automated process of bumping the versions of Gradle we test against, whilst also allowing users to pick out specific versions to test against.

Concretely, an automated process would bump versions in the `major-versions` blob, whilst leaving `extra-versions` alone.

When support for a Gradle major version is phased out, it can be just removed from this file in one fell swoop.

> [!NOTE]
> This file is global, and affects every project that is using this plugin.

### [Deprecated] Setting `gradleVersions` in the `gradleTestUtils` extension
You can set Gradle versions to test by setting `gradleVersions` property on the `gradleTestUtils` extension. For example, to test against Gradle versions `7.6.4` and `8.8`, you would do:

```groovy
gradleTestUtils {
  gradleVersions = ['7.6.4', '8.8']
}
```

> [!WARNING]
> This is deprecated because it's merges together the intent of a user wanting a _specific_ version and also the intent of testing against the latest versions of Gradle. How does an automated process know if the user wanted to test this specific version which also happens to be the latest version?

````

File: gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java [hunk 44]
```diff
@@ -207,4 +236,22 @@ private static void addGradlePluginForTestingConfigurations(Project project) {
...
+                .fileContents(testVersionsConfigPath)
+                .getAsText()
+                .map(text -> {
+                    try {
+                        return YAML_MAPPER.readValue(text, new TypeReference<>() {});
+                    } catch (JsonProcessingException e) {
+                        throw new IllegalArgumentException(e);
+                    }
+                });
+        return config.map(GradleTestVersionsConfig::allVersions);
```
felixdesouza, line 252: literally just inline this:

```java
return getProviderFactory()
                .fileContents(testVersionsConfigPath)
                .getAsText()
                .map(text -> {
                    try {
                        return YAML_MAPPER.readValue(text, new TypeReference<>() {});
                    } catch (JsonProcessingException e) {
                        throw new IllegalArgumentException(e);
                    }
                });
                .map(GradleTestVersionsConfig::allVersions);
```

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -207,4 +236,22 @@ private static void addGradlePluginForTestingConfigurations(Project project) {
...
+                getProjectLayout().getProjectDirectory().file("gradle/gradle-test-versions.yml");
+
+        Provider<GradleTestVersionsConfig> config = getProviderFactory()
+                .fileContents(testVersionsConfigPath)
+                .getAsText()
+                .map(text -> {
+                    try {
+                        return YAML_MAPPER.readValue(text, new TypeReference<>() {});
+                    } catch (JsonProcessingException e) {
+                        throw new IllegalArgumentException(e);
```
felixdesouza: let's add a helpful message here i.e. failed to deserialise test versions config file or something.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -112,8 +128,21 @@ public void execute(Task _task) {
...
+                    Set<String> gradleTestVersions = ImmutableSet.<String>builder()
+                            .addAll(versionsFromConfig)
+                            .addAll(versionsFromExtension)
+                            .build();
+                    if (gradleTestVersions.isEmpty()) {
+                        throw new IllegalStateException(
+                                "No gradleVersions were set for the PluginTestingExtension. Either set the default"
+                                        + " versions via gradle/gradle-test-versions.yml, or add overrides via the"
+                                        + " gradleVersions property");
+                    }
```
felixdesouza, line 142: I don't think we should throw here, just use the current gradle version, and don't bother with the convention in the `gradleVersions` property

README.md
felixdesouza: Also need a sentence that if nothing is set, it will use the current version of Gradle, similar to Gradle's test kit: https://docs.gradle.org/current/userguide/test_kit.html#sub:gradle-runner-gradle-version

felixdesouza:
👍

kelvinou01:
FLUPS: 
- Actually run a test in ConfigurationCacheTest

PR #290, opened 2025-11-14
Add Directory.assertThat

pkoenig10:
Same as `ProjectFile.assertThat`, but for `Directory`.

This is useful for assertions like:

```java
directory.assertThat().isNotEmptyDirectory();
```


FinlayRJW:
👍

PR #293, opened 2025-11-17
Disable configuration cache when running in debug mode

crogoz:
## After this PR

FLUP for: https://github.com/palantir/gradle-plugin-testing/pull/245#discussion_r2481068542.

Configuration cache is not available when debug is enabled (gradle/issues/25846). Adding a warning and using the DefaultGradleInvoker when cc & debug are both enabled.

==COMMIT_MSG==
Disable configuration cache when running in debug mode
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
@@ -16,17 +16,50 @@
 
 package com.palantir.gradle.testing.execution;
 
+import java.lang.management.ManagementFactory;
 import java.nio.file.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public interface GradleInvoker {
+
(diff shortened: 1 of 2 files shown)


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,17 +16,50 @@
...
 public interface GradleInvoker {
 
+    Logger log = LoggerFactory.getLogger(GradleInvoker.class);
+
     GradleInvocation withArgs(String... args);
 
     static GradleInvoker create(Path path, GradleVersion gradleVersion, boolean configurationCache) {
         DefaultGradleInvoker gradleInvoker = new DefaultGradleInvoker(path, gradleVersion);
         if (configurationCache) {
+            if (shouldRunInTestkitDebugMode()) {
```
crogoz, line 33: Tested manually by adding `systemProps idea.coverage.calculate.hits, true` property in the test task and confirmed that we are defaulting to using the DefaultGradleInvoker

FinlayRJW:
👍

PR #296, opened 2025-11-18
Add to guide based on claude feedback

esword:
## Before this PR
These are changes proposed by claude when I asked it to improve the guide to help it convert tests.  It's changes were actually more verbose so I modified them some so the guide is still useful  for humans but also help the AI. 

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/docs/testing-guide.md
+++ b/docs/testing-guide.md
@@ -163,8 +163,13 @@ All project files inherit from `ProjectFile<T>` with these methods:
 
 All methods support `String.format()` syntax for dynamic values.  The varargs overload provides better IDE support with syntax highlighting and is enforced by the `GradleTestStringFormatting` Error Prone check.
 ```java
+// ✅ CORRECT - Use varargs overload
 String version = "1.0.0";
 project.buildGradle().append("version = '%s'", version);
+
+// ❌ INCORRECT - Will fail with GradleTestStringFormatting error
+project.buildGradle().append("version = '%s'".formatted(version));
+project.buildGradle().append(String.format("version = '%s'", version));
 ```
 
 **Reading file contents:**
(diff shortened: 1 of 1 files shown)


docs/testing-guide.md
```diff
@@ -161,9 +161,15 @@ All project files inherit from `ProjectFile<T>` with these methods:
 - **`createEmpty()`** - Create an empty file
 - **`assertThat()`** - AssertJ path assertions
 
-All methods support `String.format()` syntax:
+All methods support `String.format()` syntax for dynamic values.  The varargs overload provides better IDE support with syntax highlighting and is enforced by the `GradleTestStringFormatting` Error Prone check.
 ```java
-project.buildGradle().append("version = '%s'", "1.0.0");
+// ✅ CORRECT - Use varargs overload
```
FinlayRJW: nit: I don't like the emojis but that is personal preference

docs/testing-guide.md
```diff
@@ -161,9 +161,15 @@ All project files inherit from `ProjectFile<T>` with these methods:
...
+String version = "1.0.0";
+project.buildGradle().append("version = '%s'", version);
+
+// ❌ INCORRECT - Will fail with GradleTestStringFormatting error
```
FinlayRJW: generally I'm not sure how I feel about incorrect examples, it might make sense to have these in a doc more tailored for the AI feels a bit weird having here is how not todo it in the human side

docs/testing-guide.md
esword: Ya, that's what it generated.  I'm going to guess that the "CORRECT/INCORRECT" is good enough.

docs/testing-guide.md
esword: Ya, I agree - that's why I tried to trim it down so it wasn't too much extra "fluff" for a human to read.  The issue is that I think it's good to have the CORRECT/INCORRECT next to each other for the AI.  We can try removing the incorrect example and see if it does good enough without it.

FinlayRJW: yeah I could see it being useful for AI - lets keep it in for now - I think based on what felix was saying about docs the other day we will want to have a second look over all of this anyway so can split out then

FinlayRJW:
👍

PR #297, opened 2025-11-19
[DiscoveryTestClassesTask] Task that discovers testClasses in spock and junit-jupiter test engines

crogoz:
## After this PR

FLUP for https://pl.ntr/2ye. We can now auto-discover the testClasesFiles in a repo using a `DiscoveryTestClassesTask` task. Introducing 2 tasks: 
1. `discoverGradlePluginTests` discovers GradlePluginTests annotated test classes. This will be used by the Configuration cache enablement migration to discover the tests that need to be annotated with `@DisabledConfigurationCache`.
2. `discoverNebulaTestClassesToMigrate` discovers the groovy test classes that extend `IntegrationSpec`, `IntegrationTeskKitSpec` excluding the `ConfigurationCacheSpec` tests. This will be used by the excavator to discover the groovy tests that need to be migrated to the new junit framework.

==COMMIT_MSG==
[DiscoveryTestClassesTask] Task that discovers testClasses in spock and junit-jupiter test engines
==COMMIT_MSG==


discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
```diff
@@ -0,0 +1,75 @@
...
+    static boolean isSubclassOf(Class<?> originalClass, String targetClassName) {
+        Class<?> clazz = originalClass;
+        while (clazz != null && !clazz.equals(Object.class)) {
+            if (clazz.getName().equals(targetClassName)) {
+                return true;
+            }
+            clazz = clazz.getSuperclass();
+        }
+        return false;
+    }
```
felixdesouza: isn't this just an `instanceof` check?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoveryTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+        String classType = getTestClassType().get();
+        if (classType.equals("java")) {
+            return "junit-jupiter";
+        } else if (classType.equals("groovy")) {
+            return "spock";
+        }
+        throw new RuntimeException(String.format("Invalid classType %s", classType));
+    }
+
+    private boolean pathMatches(Path path, Set<String> testClasses) {
```
felixdesouza: nit: static

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+    public abstract RegularFileProperty getOutputFile();
+
+    @OutputFile
+    public abstract RegularFileProperty getDiscoveredTestsFile();
+
+    @Inject
+    public abstract ProjectLayout getProjectLayout();
+
+    @Inject
+    public abstract ProviderFactory getProviderFactory();
```
felixdesouza, line 76: can we group the inputs together and move the `@Inject`s below them pls

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoveryTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+    public abstract Property<File> getParentPath();
+
+    @Input
+    public abstract ListProperty<String> getExtraArguments();
+
+    @Input
+    public abstract Property<String> getTestClassType();
+
+    public DiscoveryTestClassesTask() {
+        setClasspath(getProject().files(getRuntimeClasspath(), getTestClasspath()));
```
felixdesouza: I think you'd get dinged on the `getProject` call? in any case perhaps just use the `objectFactory`?

I even think you can do constructor injection instead of method injection, but idk if we like that style, up to you

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoveryTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+                                getProjectLayout().getBuildDirectory(),
+                                (type, buildDir) ->
+                                        buildDir.dir("tests-discovery").file(String.format("%s-test-classes", type))));
+        getOutputFile()
+                .convention(getProviderFactory()
+                        .zip(
+                                getTestClassType(),
+                                getProjectLayout().getBuildDirectory(),
+                                (type, buildDir) -> buildDir.dir("tests-discovery")
+                                        .file(String.format("%s-test-classes-paths", type))));
```
felixdesouza: can we do instead:

```java
Provider<Directory> outputRootDirectory = getProviderFactory().zip(
  getProjectLayout().getBuildDirectory(), 
  getTestClassType(),
  (buildDirectory, testType) -> buildDirectory.dir(String.format("tests-discovery/%s", testType)));

getDiscoveredTestsFile().convention(outputRootDirectory.map(outputRoot -> outputRoot.file("test-classes")));
getOutputFile().convention(outputRootDirectory.map(outputRoot -> outputRoot.file("test-classes-paths")));
```

I think it's a bit nicer not having to parse file names for types of things, and it's all together nicely in a directory.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoveryTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+        getMainClass().set("com.palantir.gradle.plugintesting.DiscoverTestsMain");
+        getArgumentProviders().add(this::getArguments);
+
+        doLast(new Action<Task>() {
+            @Override
+            public void execute(Task _task) {
+                try {
+                    Path parentPath = getParentPath().get().toPath();
+                    Set<String> discoveredTestClassesPaths =
+                            Files.readAllLines(getDiscoveredTestsFile()
```
felixdesouza: nit, move `getDiscoveredTestsFile().getAsFile().get().toPath()` on a separate line/variable

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+                    Path parentPath = getParentPath().get().toPath();
+                    Set<String> discoveredTestClassesPaths =
+                            Files.readAllLines(getDiscoveredTestsFile()
+                                            .getAsFile()
+                                            .get()
+                                            .toPath())
+                                    .stream()
+                                    .map(this::toFilePath)
+                                    .collect(Collectors.toSet());
+                    Set<String> discoveredTestFiles = getTestSourceFiles().getAsFileTree().getFiles().stream()
```
felixdesouza, line 105: I'd be a bit curious about the validation here, especially for junit types, we're starting to use static inner classes via nested tests and they'd have a different classname I believe? I think it should be completely fine with spock though

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -49,20 +49,23 @@
 import org.gradle.plugin.devel.plugins.JavaGradlePluginPlugin;
 import org.gradle.plugin.devel.tasks.PluginUnderTestMetadata;
 import org.gradle.util.GradleVersion;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
felixdesouza: let's use the gradle imports instead:

```java
import org.gradle.api.logging.Logger;
import org.gradle.api.logging.Logging;
```

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -158,9 +161,40 @@ public void execute(Task _task) {
...
+                            "subClassesOf",
+                            "--include",
+                            "nebula.test.IntegrationSpec,nebula.test.IntegrationTestKitSpec",
+                            "--exclude",
+                            "com.palantir.gradle.plugintesting.ConfigurationCacheSpec"));
+        });
+
+        project.getTasks().withType(DiscoveryTestClassesTask.class, task -> {
+            SourceSetContainer sourceSetContainer = project.getExtensions().getByType(SourceSetContainer.class);
+            SourceSet testSourceSet = sourceSetContainer.getByName(SourceSet.TEST_SOURCE_SET_NAME);
```
felixdesouza, line 190: I think we want to capture `integrationTest` as well? That said, there aren't that many of them internally.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -158,9 +161,40 @@ public void execute(Task _task) {
...
+            task.getTestClasspath().setFrom(testSourceSet.getRuntimeClasspath());
+            task.getRuntimeClasspath().setFrom(project.getConfigurations().getByName("runtimeClasspath"));
```
felixdesouza: how are you capturing the classes that are `discover-tests-cli` ?

I think it's a little weird to put the `runtimeClasspath` on here.

We should probably make a configuration that contains just the `discovery-tests-cli` module and union that with the test classpath. 

We should probably lock it as well since we wouldn't want it to clash - although we can probably hold off on this part until we see any issues.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -158,9 +161,40 @@ public void execute(Task _task) {
...
+            task.getTestSourceFiles().setFrom(testSourceSet.getAllSource());
+            task.getParentPath().set(project.getRootProject().getProjectDir());
```
felixdesouza: the parent isn't guaranteed to be the root project? can't we just do `project.getParent()` here instead?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -158,9 +161,40 @@ public void execute(Task _task) {
             });
         });
 
+        addTestClassesDiscoveryTasks(project);
+
         addGradlePluginForTestingConfigurations(project);
     }
 
+    private static void addTestClassesDiscoveryTasks(Project project) {
+        project.getTasks().register("discoverGradlePluginTests", DiscoveryTestClassesTask.class, task -> {
```
felixdesouza: these look really neat!

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+                            .map(File::toPath)
+                            .filter(path -> pathMatches(path, discoveredTestClassesPaths))
+                            .map(path -> parentPath.relativize(path).toString())
```
felixdesouza, line 108: is the idea here that you'll get a relative path relative to the root directory then? if so, we should call it `rootPath` instead of parent path.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+                            .collect(Collectors.toSet());
+                    if (discoveredTestFiles.size() != discoveredTestClassesPaths.size()) {
+                        throw new RuntimeException(String.format(
+                                "Could not find all test source classes for the discovered test classes. Discovered"
+                                        + " tests: %s, but only mapped the following classes in the repo: %s",
+                                discoveredTestClassesPaths, discoveredTestFiles));
+                    }
+                    Files.writeString(
+                            getOutputFile().getAsFile().get().toPath(), String.join("\n", discoveredTestFiles));
```
felixdesouza: I think you can just do `Files.write(getOutputFile().getAsFile().get().toPath(), discoveredTestFiles)`

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoveryTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+                    Set<String> discoveredTestFiles = getTestSourceFiles().getAsFileTree().getFiles().stream()
+                            .map(File::toPath)
+                            .filter(path -> pathMatches(path, discoveredTestClassesPaths))
+                            .map(path -> parentPath.relativize(path).toString())
+                            .collect(Collectors.toSet());
+                    if (discoveredTestFiles.size() != discoveredTestClassesPaths.size()) {
+                        throw new RuntimeException(String.format(
+                                "Could not find all test source classes for the discovered test classes. Discovered"
+                                        + " tests: %s, but only mapped the following classes in the repo: %s",
+                                discoveredTestClassesPaths, discoveredTestFiles));
```
felixdesouza: can we output the difference i.e. what was missing? Can either be in addition to or in place of, up to you

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoveryTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.InputFiles;
+import org.gradle.api.tasks.JavaExec;
+import org.gradle.api.tasks.OutputFile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class DiscoveryTestClassesTask extends JavaExec {
+
+    private static final Logger log = LoggerFactory.getLogger(DiscoveryTestClassesTask.class);
```
felixdesouza: same comment re gradle logger

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoveryTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+import org.gradle.api.provider.Property;
+import org.gradle.api.provider.ProviderFactory;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.InputFiles;
+import org.gradle.api.tasks.JavaExec;
+import org.gradle.api.tasks.OutputFile;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class DiscoveryTestClassesTask extends JavaExec {
```
esword: nit: name - DiscoverTestClassesTask?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -158,9 +161,40 @@ public void execute(Task _task) {
...
+            task.getTestClassType().set("java");
+            task.getExtraArguments()
+                    .addAll(List.of(
+                            "withAnnotations", "--include", "com.palantir.gradle.testing.junit.GradlePluginTests"));
+        });
+
+        project.getTasks().register("discoverGroovyTestClassesToMigrate", DiscoveryTestClassesTask.class, task -> {
```
felixdesouza: perhaps let's call it `discoverNebulaTestClassesToMigrate` ?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -158,9 +161,40 @@ public void execute(Task _task) {
...
+            task.getTestClassType().set("groovy");
+            task.getExtraArguments()
+                    .addAll(List.of(
+                            "subClassesOf",
+                            "--include",
+                            "nebula.test.IntegrationSpec,nebula.test.IntegrationTestKitSpec",
```
felixdesouza, line 183: is there a blocker on migrating the older school `IntegrationSpec` classes?

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/TestClassesDiscoverer.java
```diff
@@ -0,0 +1,105 @@
...
+        LauncherDiscoveryRequest discoveryRequest = LauncherDiscoveryRequestBuilder.request()
+                .selectors(DiscoverySelectors.selectClasspathRoots(Arrays.stream(classPath.split(File.pathSeparator))
+                        .map(Paths::get)
+                        .collect(Collectors.toSet())))
+                .filters(getFilter())
+                .build();
+
+        LauncherConfig launcherConfig = LauncherConfig.builder()
+                .enableTestEngineAutoRegistration(false)
+                .addTestEngines(testClassesCommand.getTestEngine())
```
felixdesouza: I wonder whether you can avoid a dependency on the individual test engines themselves by using the auto registration. Then you could just do an engine filter in the discovery request above:

```java
        LauncherDiscoveryRequest discoveryRequest = LauncherDiscoveryRequestBuilder.request()
                .selectors(DiscoverySelectors.selectClasspathRoots(Arrays.stream(classPath.split(File.pathSeparator))
                        .map(Paths::get)
                        .collect(Collectors.toSet())))
                
                .filters(getFilter(), EngineFilter.includeEngines(testClassesCommand.getTestEngine()))
                .build();
```

of course `getTestEngine` will just be a string now.

The main benefit here is that you have fewer dependencies that may clash with the actual test classpath and it's just whatever the test classpath contributes.

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestsMain.java
```diff
@@ -0,0 +1,38 @@
...
+import java.nio.file.Path;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+
+@Command(name = "discover", subcommands = TestClassesCommand.class)
+public final class DiscoverTestsMain {
+
+    @Option(names = "--output", description = "Output")
+    private String output;
```
felixdesouza: I think you can just use `Path` directly, it's [supported out of the box](https://picocli.info/#_built_in_types)!

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/TestClassesDiscoverer.java
```diff
@@ -0,0 +1,105 @@
...
+                            .map(TestIdentifier::getSource)
+                            .flatMap(test -> test.stream()
+                                    .map(TestClassesDiscoverer::getTestClassFromSource)
+                                    .map(clazz -> clazz.getName())))
+                    .collect(Collectors.toSet());
+        }
+    }
+
+    @Override
+    public final Integer call() throws Exception {
```
felixdesouza: is there any point in making it `Callable` ? we're implicitly relying on picocli returning a non zero exit code in the face of an exception. We might as well just implement `Runnable` instead?

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/TestClassesDiscoverer.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
+import org.junit.platform.launcher.core.LauncherFactory;
+import picocli.CommandLine.ParentCommand;
+
+public abstract class TestClassesDiscoverer implements Callable<Integer> {
+
+    @ParentCommand
+    private TestClassesCommand testClassesCommand;
+
+    abstract Filter<?> getFilter();
```
felixdesouza: nit: protected

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/TestClassesDiscoverer.java
```diff
@@ -0,0 +1,105 @@
...
+
+    final Set<String> getDiscoveredClassNames() {
```
felixdesouza: nit: private

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
```diff
@@ -0,0 +1,159 @@
...
+
+            private String toFilePath(String line) {
+                return line.replace('.', '/') + "." + getTestClassType().get();
+            }
+        });
+    }
+
+    private List<String> getArguments() {
+        return Stream.concat(
+                        Stream.of(
```
felixdesouza, line 133: I'm surprised you don't have to put `discover` as the first argument here. Nothing actionable, just surprised is all

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
```diff
@@ -199,4 +210,104 @@ void test() throws Exception {
...
+                .isEqualTo("src/test/java/test/GradlePluginTestClass.java");
+    }
+
+    @Test
+    void groovy_tests_ready_for_migration_are_discovered(GradleInvoker gradle, RootProject rootProject)
+            throws IOException {
+        Directory groovyDir =
+                rootProject.testSourceSet().srcDir("groovy").directory("test").createDirectories();
+
+        Map<String, String> testNameToClass = Map.of(
```
felixdesouza, line 265: could we not do a parameterised test here?

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
```diff
@@ -199,4 +210,104 @@ void test() throws Exception {
...
+        gradle.withArgs("discoverGroovyTestClassesToMigrate").buildsSuccessfully();
+        rootProject
+                .buildDir()
+                .directory("tests-discovery")
+                .file("groovy-test-classes-paths")
+                .assertThat()
+                .content()
+                .hasLineCount(2)
+                .contains("src/test/groovy/test/NebulaIntegrationTest.groovy")
+                .contains("src/test/groovy/test/NebulaIntegrationTestKitSpec.groovy");
```
felixdesouza: let's just do `containsExactlyInAnyOrder(...)`

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/TestClassesDiscoverer.java
```diff
@@ -0,0 +1,105 @@
...
+    static Class<?> getTestClassFromSource(TestSource testSource) {
+        if (testSource instanceof ClassSource classSource) {
+            return classSource.getJavaClass();
+        } else if (testSource instanceof MethodSource methodSource) {
+            return methodSource.getJavaClass();
+        }
+        throw new IllegalStateException(String.format("Unexpected testSource type %s", testSource));
+    }
+
+    private static void writeOutput(Path output, Set<String> classes) {
```
felixdesouza, line 90: this task may frequently get out of date due to arbitrary iteration order. We should sort it and then write it out

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/TestClassesDiscoverer.java
```diff
@@ -0,0 +1,105 @@
...
+        try {
+            Files.writeString(
```
felixdesouza: We can just do `Files.write(output, classes.stream().sorted().toList(), StandardOpenOption.TRUNCATE_EXISTING, StandardOpenOption.WRITE, StandardOpenOption.CREATE)`

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/TestClassesDiscoverer.java
```diff
@@ -0,0 +1,105 @@
...
+                    output,
+                    String.join("\n", classes),
+                    StandardOpenOption.TRUNCATE_EXISTING,
+                    StandardOpenOption.WRITE,
+                    StandardOpenOption.CREATE);
+        } catch (IOException e) {
+            throw new UncheckedIOException(String.format("Failed to write the output to file %s", output), e);
```
felixdesouza: Something a bit more specific: "failed to write test discovery output to file" bla bla

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
crogoz: yes, but I don't think you can pass the class name to the instanceOf check :(

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
crogoz: I actually need the relative path to the rootProject, I'll rename getParentPath -> getRootPath

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
felixdesouza: hmm, but in the annotation command, you're doing a `Class.forName`, so could you not just do that here as well?

felixdesouza: if the target class you're checking is not on the classpath, that's fine, then it just means that it can't be implemented by the test class either! So you could do a prefilter of the `targetClasses` and then check that instead with `instanceOf` or whatever the method is on `Class`

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
crogoz: I think the toString representation of the discoveredTestFiles (which is a set) will include the square brackets eg. [/some/path,/other/path]

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
felixdesouza: actually, you want `Class#isAssignableFrom` since you're not actually dealing with objects here

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
felixdesouza: yeah, I got to the same conclusion at the point of usage

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
felixdesouza: `Files.write` takes in an `Iterable<? extends CharSequence>` so should be okay?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
crogoz: I am including `nebula.test.IntegrationSpec` in the migration logic. is there a different one I should look at ?

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/TestClassesDiscoverer.java
crogoz: nice! this worked

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
felixdesouza: lmao i cannot read, ignore me

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
crogoz: `discover` is the placeholder for the Main class, I don't think it would work adding `discover` here as I am calling the Main class with it's argument (which int this case is the `testClasses` subcommand)

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
crogoz: I wanted to also make sure I can see the output makes sense when discovering multiple testClasses

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
crogoz: I think we might want to migrate spock test classes as well in the future - eg. Specification ? wdyt ?

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
felixdesouza: ah okay, makes sense, I misread and thought there were gradle invocations being run

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
```diff
@@ -199,4 +210,104 @@ void test() throws Exception {
...
+                // ignored because it doesn't extend an allowlisted class
+                "IgnoredSpecification",
+                "spock.lang.Specification");
+        testNameToClass.forEach((testName, importClass) -> {
+            try {
+                List<String> parts = Splitter.on('.').splitToList(importClass);
+
+                String classToExtend = parts.get(parts.size() - 1);
+                Files.writeString(
+                        groovyDir.file(String.format("%s.groovy", testName)).path(),
```
felixdesouza: don't we have a helper to write files here?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
crogoz: I don't think I've seen nested tests, when would those make sense ? esp as static inner test classes ? 
Do you have some examples for this ? I can add a few test cases for this.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
felixdesouza: It's hard to say, for example the `ProjectSpec` and a bunch of others don't have an equivalent migration, they don't even run with Gradle or maybe there is a different thing we can do in testkit idk. Some things that are just `Specification` and no link to `IntegrationSpec` it's also unclear what/how they should be migrated.

I would not be opposed to having a generic "discover spock classes" thing that we can use to burn down stuff. It's not clear to me we can reuse the same migration logic for each thing + I *think* we'd need to specify what sort of base class it is or what sort of migration it'd require perhaps idk. Wdyt?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
felixdesouza: It's actually quite nice, it allows for better organisation in a test class. [Here's](https://github.com/palantir/gradle-baseline/blob/develop/gradle-baseline-java/src/test/java/com/palantir/baseline/BaselineJavaVersionIntegrationTest.java#L33-L33) and example from gradle-baseline that was added recently. 

Plus I don't think there'd be anything stopping people from wanting to use it, and you wouldn't want this task to blow up if they started using it but still had groovy classes to migrate

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
crogoz: I actually don't need this anymore - as the "testImplementation" of the project already has the "discovery-tests-cli" here: https://github.com/palantir/gradle-plugin-testing/pull/297/files#diff-ca7ede7b34d2580d6fdb25b3707c5007eb08095357c26ab8c22bcb48236f2b85R232. Let me know if you prefer to have a separate config for this though

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
crogoz: I don't think this project is wiring in other places other sourceSets besides the test one. Should we add it as a FLUP ?

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
crogoz: you are right! changed it [here](https://github.com/palantir/gradle-plugin-testing/pull/297/commits/0e5ebcbe2be4d1fc6dc2b319f6b463c34d6ff3e9)

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
crogoz: ah, I see. Checked now and the code is working - it shows correctly the file classes even if we have nested classes. And this is because we are using `TestIdentifier::getSource` which will be the same for the nested class and the top-level annotated class.

I think this should be fine, as for the migration for the junit tests I might only need them atm for enabling the configuration cache, and I think getting the top-level annotated class might be enough.

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
crogoz: Makes sense, I modified it to `discoverNebulaTestClassesToMigrate`

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/DiscoverTestClassesTask.java
crogoz: oh, you are right! changed it

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
crogoz: changed it in https://github.com/palantir/gradle-plugin-testing/pull/297/commits/173503621ad0e5189aa65e14c9fcfbe8ed403f7e

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
```diff
@@ -218,14 +217,23 @@ void gradlePluginTests_are_discovered(GradleInvoker gradle, RootProject rootProj
 
             import com.palantir.gradle.testing.junit.GradlePluginTests;
             import java.nio.file.Files;
-            import org.junit.jupiter.api.Test;
+            import org.junit.jupiter.api.Nested;import org.junit.jupiter.api.Test;
```
felixdesouza: nit: 
```suggestion
            import org.junit.jupiter.api.Nested;
            import org.junit.jupiter.api.Test;
```

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
```diff
@@ -243,17 +251,16 @@ void my_junit_test() {
...
 
         gradle.withArgs("discoverGradlePluginTests", "--info").buildsSuccessfully();
-        rootProject
-                .buildDir()
-                .directory("tests-discovery")
-                .file("java-test-classes-paths")
-                .assertThat()
-                .content()
-                .isEqualTo("src/test/java/test/GradlePluginTestClass.java");
+        assertThat(Files.readAllLines(rootProject
```
felixdesouza: nit: let's make a static method for:

```java
Files.readAllLines(rootProject.buildDir().directory("tests-discovery/java").file("test-classes-paths").path());
```

e.g. `readTestClassesPaths` or something

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
```diff
@@ -28,24 +30,27 @@
 public final class SubClassesOfCommand extends TestClassesDiscoverer {
 
     @Option(names = "--include", split = ",", description = "List of subClasses to include")
-    private List<String> subClasses;
+    private List<String> classNames;
 
     @Option(names = "--exclude", split = ",", description = "List of subClasses to exclude")
```
felixdesouza: nit: let's update the description with the new change

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/WithAnnotationsCommand.java
```diff
@@ -0,0 +1,68 @@
...
+                        .orElseGet(() -> FilterResult.excluded("Not a class source"));
+            }
+        };
+    }
+
+    private static boolean hasAnyClassAnnotations(Class<?> clazz, List<String> annotations) {
+        return annotations.stream().anyMatch(annotation -> hasClassAnnotation(clazz, annotation));
+    }
+
+    private static boolean hasClassAnnotation(Class<?> clazz, String annotationName) {
```
felixdesouza: something I forgot to mention the first time. We should use `AnnotationSupport#isAnnotated` it's in the junit-platform-commons jar

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/WithAnnotationsCommand.java
```diff
@@ -0,0 +1,68 @@
...
+
+@Command(name = "withAnnotations", description = "Test classes options")
+public final class WithAnnotationsCommand extends TestClassesDiscoverer {
+
+    @Option(names = "--include", split = ",", description = "List of testClasses with annotations to include")
+    private List<String> includedAnnotations;
+
+    @Override
+    protected Filter<?> getFilter() {
+        return new PostDiscoveryFilter() {
```
felixdesouza, line 42: can we convert the list of `includedAnnotations` to `List<Class<? extends Annotation>>`  here, and then reference that below?

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
felixdesouza: yeah, let's think about it as a flup, it will just be less coverage, so nothing is actually wrong per se

discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/WithAnnotationsCommand.java
felixdesouza: we won't pay the cost of loading the class each time, since it's cached within the classloader, but there's still a bit of code around that which scales with the number of tests a project has.

felixdesouza: do you want to just filter here rather than throw? then you don't fail when you try to run this command on a classpath that doesn't have the annotations you care about. That said, I'm not sure if it will happen in practice

felixdesouza:
👍  let's gooo 🚀

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
dmopsickcw: The inferred-used result has been removed in the latest revision. An unexpired invalid token can result from consumption, tampering, or other user-state invalidation, so we now return `InvalidToken` rather than presenting an unreliable specific result. I also updated the PR description to match this behavior.

PR #305, opened 2025-11-26
don't check if configuration time error that isn't related to configuration cache error

FinlayRJW:
## Before this PR
We would always check the configuration cache was stored and that dry-run was succesful

## After this PR
after https://github.com/palantir/gradle-shadow-jar/pull/895 we noticed we dont want to check the store and dry-run if the configuration time error that isn't related to configuration cache error

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTests.java
+++ b/gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTests.java
@@ -183,17 +183,12 @@ class ConfigurationCacheTests {
     @Test
     void configuration_time_errors_do_not_run_dry_run(GradleInvoker invoker, RootProject rootProject) {
         rootProject.buildGradle().append("""
-            project.afterEvaluate(_p -> {
-                if (!project.getRootProject().getPlugins().hasPlugin("randomPlugin")) {
-                    throw new IllegalStateException("Missing a random plugin");
-                }
-            });
+            afterEvaluate {
+                throw new IllegalStateException("Configuration time issue");
+            }
             """);
 
         InvocationResult result = invoker.withArgs("help").buildsWithFailure();
-        result.assertThat()
-                .output()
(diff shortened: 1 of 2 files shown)


gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ConfigurationCacheTests.java
```diff
@@ -179,4 +179,21 @@ void run() {
...
     }
+
+    @Test
+    void configuration_time_errors_do_not_run_dry_run(GradleInvoker invoker, RootProject rootProject) {
+        rootProject.buildGradle().append("""
+            project.afterEvaluate(_p -> {
+                if (!project.getRootProject().getPlugins().hasPlugin("randomPlugin")) {
+                    throw new IllegalStateException("Missing a random plugin");
+                }
+            });
```
CRogers: I would just simplify this test to be `throw new IllegalStateException("Configuration time issue");`. No need for the afterEvaluate or plugins etc, makes the test more complicated than it needs to be imo.

FinlayRJW: throwing immediatly is a `org.gradle.api.GradleScriptException` error rather than a `ProjectConfigurationException` - should we be checking both?

FinlayRJW: I think we want 
```java
           afterEvaluate {
                throw new IllegalStateException("Configuration time issue");
            }
```

as without after evaluate we fail at script complication?

CRogers: We should absolutely be checking both assuming the `GradleScriptException` also causes the `--dry-run` step to go wrong, we need a test case for each

CRogers (approved):

PR #309, opened 2025-12-02
Fix order of `DisabledConfigurationCache` annotation

crogoz:
## Before this PR

When `@DisabledConfigurationCache` was set before the `@GradlePluginTests`, the CC store value wouldn't be overridden. (internal example [here](https://pl.ntr/2ym))

## After this PR
CC is disabled when `@DisabledConfigurationCache` is applied regardless the order of annotations

==COMMIT_MSG==
Fix order of `DisabledConfigurationCache` annotation
==COMMIT_MSG==


gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/DisabledConfigurationCacheTestBefore.java
```diff
@@ -0,0 +1,37 @@
...
+package com.palantir.gradle.testing.junit;
+
+import com.palantir.gradle.testing.execution.GradleInvoker;
+import com.palantir.gradle.testing.execution.InvocationResult;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.Test;
+
+@DisabledConfigurationCache
+@GradlePluginTests
+public class DisabledConfigurationCacheTestBefore {
```
felixdesouza: this is a bit of a weird test in that it's based on the repo's set up rather than a test fixture, and it took me a minute to understand the importance of this test

felixdesouza: I would have probably put the cc test's as fixtures and then load them in builds under test as I don't think this repo's set up should influence how the tests run + it makes it clearer what you're trying to assert on

esword (approved):

PR #310, opened 2025-12-02
Add `discoverGradlePluginTestsWithDisabledConfigurationCache`  task

crogoz:
## After this PR
Needed by the internal excavator that does the migration to CC. The excavator needs to generate all the test files that have `@DisabledConfigurationCache` annotation such that for each one of them it can try to remove the annotation & check if the tests are compatible with CC.

==COMMIT_MSG==
Add `discoverGradlePluginTestsWithDisabledConfigurationCache`  task
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
+++ b/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
@@ -167,12 +167,18 @@ public abstract class PluginTestingPlugin implements Plugin<Project> {
     }
 
     private static void addTestClassesDiscoveryTasks(Project project) {
-        project.getTasks().register("discoverGradlePluginTests", DiscoverTestClassesTask.class, task -> {
-            task.getTestClassType().set("java");
-            task.getExtraArguments()
-                    .addAll(List.of(
-                            "withAnnotations", "--include", "com.palantir.gradle.testing.junit.GradlePluginTests"));
-        });
+        project.getTasks()
+                .register(
+                        "discoverDisabledConfigurationCacheGradlePluginTests", DiscoverTestClassesTask.class, task -> {
+                            task.getTestClassType().set("java");
+                            task.getExtraArguments()
+                                    .addAll(List.of(
+                                            "withAnnotations",
+                                            "--include",
+                                            "com.palantir.gradle.testing.junit.GradlePluginTests",
+                                            "--exclude",
+                                            "com.palantir.gradle.testing.junit.DisabledConfigurationCache"));
+                        });
 
         project.getTasks().register("discoverNebulaTestClassesToMigrate", DiscoverTestClassesTask.class, task -> {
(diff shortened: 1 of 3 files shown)


gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -167,12 +167,18 @@ public void execute(Task _task) {
...
     private static void addTestClassesDiscoveryTasks(Project project) {
-        project.getTasks().register("discoverGradlePluginTests", DiscoverTestClassesTask.class, task -> {
-            task.getTestClassType().set("java");
-            task.getExtraArguments()
-                    .addAll(List.of(
-                            "withAnnotations", "--include", "com.palantir.gradle.testing.junit.GradlePluginTests"));
-        });
+        project.getTasks()
+                .register(
+                        "discoverDisabledConfigurationCacheGradlePluginTests", DiscoverTestClassesTask.class, task -> {
```
esword: My $.02 - I'd put the main part of what the task does first in the name, then the "qualifier" part of it - discoverGradlePluginTestsWithDisabledConfigurationCache

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
```diff
@@ -238,6 +238,34 @@ void nested_test() throws Exception {
...
+        rootProject.testSourceSet().java().writeClass("""
+            package test;
+
+            import com.palantir.gradle.testing.junit.GradlePluginTests;
+            import com.palantir.gradle.testing.junit.DisabledConfigurationCache;
+            import java.nio.file.Files;
+            import org.junit.jupiter.api.Nested;
+            import org.junit.jupiter.api.Test;
+
+            @DisabledConfigurationCache
```
esword: do you want to put DisabledConfigurationCache after GradlePluginTests since you had another PR that positions it that way?

gradle-plugin-testing/src/test/java/com/palantir/gradle/plugintesting/PluginTestingJunitPluginTest.java
```diff
@@ -251,7 +279,7 @@ void my_junit_test() {
             }
             """);
 
-        gradle.withArgs("discoverGradlePluginTests", "--info").buildsSuccessfully();
+        gradle.withArgs("discoverDisabledConfigurationCacheGradlePluginTests").buildsSuccessfully();
```
esword: are you doing anything with that new class you are creating?

crogoz: no, I just wanted to make sure it is not discovered by the task because it has the @DisabledCC annotaiton

esword (approved):

PR #312, opened 2025-12-02
Fix `discoverGradlePluginTestsWithDisabledConfigurationCache`

crogoz:
## After this PR

`discoverGradlePluginTestsWithDisabledConfigurationCache` needs to check for the `@DisabledConfigurationCache` annotation only.
==COMMIT_MSG==
Fix `discoverGradlePluginTestsWithDisabledConfigurationCache`
==COMMIT_MSG==

### DIFF
--- a/discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
+++ b/discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
@@ -29,10 +29,10 @@ import picocli.CommandLine.Option;
 @Command(name = "subClassesOf", description = "Test classes options")
 public final class SubClassesOfCommand extends TestClassesDiscoverer {
 
-    @Option(names = "--include", split = ",", description = "List of sub-class names to include")
+    @Option(names = "--include", split = ",", description = "List of sub-class names to include", defaultValue = "")
     private List<String> classNames;
 
-    @Option(names = "--exclude", split = ",", description = "List of sub-class names to exclude")
+    @Option(names = "--exclude", split = ",", description = "List of sub-class names to exclude", defaultValue = "")
     private List<String> excludedClassNames;
 
     @Override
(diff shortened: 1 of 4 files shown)


discover-tests-cli/src/main/java/com/palantir/gradle/plugintesting/SubClassesOfCommand.java
```diff
@@ -29,10 +29,10 @@
 @Command(name = "subClassesOf", description = "Test classes options")
 public final class SubClassesOfCommand extends TestClassesDiscoverer {
 
-    @Option(names = "--include", split = ",", description = "List of sub-class names to include")
+    @Option(names = "--include", split = ",", description = "List of sub-class names to include", defaultValue = "")
     private List<String> classNames;
 
-    @Option(names = "--exclude", split = ",", description = "List of sub-class names to exclude")
+    @Option(names = "--exclude", split = ",", description = "List of sub-class names to exclude", defaultValue = "")
```
crogoz, line 35: in case --exclude/--include is not specified the defaultValue would be null, which will later fail when we are trying to extract the classes: https://github.com/palantir/gradle-plugin-testing/pull/312/files#diff-5bbff74157d9b58b51f10fa8cd4406e907f2cb2b084e22426a37734a0e7874bfR42

gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -177,8 +177,6 @@ private static void addTestClassesDiscoveryTasks(Project project) {
                                     .addAll(List.of(
                                             "withAnnotations",
                                             "--include",
-                                            "com.palantir.gradle.testing.junit.GradlePluginTests",
```
crogoz, line 180: we need the inclusion of the `DisabledConfigurationCache` - see excavator logic [here](https://pl.ntr/2yo)

esword (approved):

PR #316, opened 2025-12-09
handle multiple buildscripts when repositioning plugins block

FinlayRJW:
## Before this PR

We could only handle a single buildscript block. When positioning the plugins block, we would place it after the first buildscript found. This was problematic because:

1. Multiple buildscript blocks are valid Gradle - Gradle allows multiple buildscript blocks in a single build file, and they are evaluated in order
2. Common test patterns broke - A user might have a helper like standardBuildfile(rootProject) that sets up common configuration including a buildscript, then later .append("""buildscript { ... }""") to add test-specific classpath dependencies
3. Workaround was manual editing - Previously, users had to manually use .edit() to merge multiple buildscript blocks into one to avoid the plugins block being incorrectly positioned between them

## After this PR

Now we correctly handle multiple buildscript blocks by:
- Extracting all buildscript blocks from the file
- Moving them all to the top of the file
- Positioning the plugins block after all buildscripts
- Placing the remaining content after the plugins block

This means patterns like this now work correctly:

```java
standardBuildfile(rootProject).append("""
    buildscript {
        dependencies {
            classpath 'com.example:plugin:1.0'
        }
    }
    """);
rootProject.buildGradle().plugins().add("java");
```

The result will have both buildscript blocks at the top, followed by the plugins block.

==COMMIT_MSG==
Handle multiple buildscript blocks when positioning plugins

Previously, adding plugins to a file with multiple buildscript blocks
would incorrectly position the plugins block after only the first
buildscript. Now all buildscript blocks are collected and moved to the
top, with the plugins block positioned after all of them.
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
@@ -16,7 +16,7 @@
 
 package com.palantir.gradle.testing.files.gradle;
 
-import java.util.ArrayList;
+import com.google.common.collect.Lists;
 import java.util.List;
 import java.util.Optional;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -89,16 +89,37 @@ private String insertPluginsBlock(String text, List<Plugin> plugins) {
...
+            int openBraceIndex = matcher.end() - 1;
+
+            // Find matching closing brace
+            Optional<Integer> endIndex = findMatchingBrace(remainingText, openBraceIndex);
+            if (endIndex.isEmpty()) {
+                break;
+            }
+
+            buildscriptBlocks.add(remainingText.substring(startIndex, endIndex.get()));
+            remainingText = remainingText.substring(0, startIndex) + remainingText.substring(endIndex.get());
```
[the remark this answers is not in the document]
FinlayRJW: dont need / want to preserve spacing

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -137,14 +156,9 @@ static String repositionPluginsBlock(String text) {
...
-    private static OptionalInt findBuildScriptBlockEnd(String text) {
-        Matcher matcher = BUILDSCRIPT_PATTERN.matcher(text);
-        if (!matcher.find()) {
-            return OptionalInt.empty();
-        }
-        int openBraceIndex = matcher.end() - 1; // position of '{'
+    private record Range(int start, int end) {}
 
-        // Match braces
+    private static Optional<Integer> findMatchingBrace(String text, int openBraceIndex) {
```
esword, line 159: why change this from OptionalInt

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
```diff
@@ -137,14 +156,9 @@ static String repositionPluginsBlock(String text) {
         return repositionPluginsBlockWithContent(textWithoutPluginsBlock, pluginsBlock);
     }
 
-    private static OptionalInt findBuildScriptBlockEnd(String text) {
-        Matcher matcher = BUILDSCRIPT_PATTERN.matcher(text);
-        if (!matcher.find()) {
-            return OptionalInt.empty();
-        }
-        int openBraceIndex = matcher.end() - 1; // position of '{'
+    private record Range(int start, int end) {}
```
FinlayRJW: whoops need to delete

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/Plugins.java
FinlayRJW: I don't think you can map on `OptionalInt` so you would need todo:

```java
.map(match -> {
    OptionalInt endIndex = findMatchingBrace(textWithoutPluginsBlock, match.end() - 1);
    return endIndex.isPresent()
            ? Optional.of(new int[] {match.start(), endIndex.getAsInt()})
            : Optional.<int[]>empty();
})
```

instead of

```java
.map(match -> findMatchingBrace(textWithoutPluginsBlock, match.end() - 1)
                        .map(endIndex -> new int[] {match.start(), endIndex}))
```

FinlayRJW:
We don't actually need to first `BuildScript` in the test I was migrating so will close for now

PR #319, opened 2025-12-15
Wrap `@GradlePluginTests` TestKit build outputs in nice color-coded boxes

kelvinou01:
## Before this PR

In a `@GradlePluginTest`, the build outputs of the 1. test driver "outer" build 2. multiple "inner" testkit builds[^1] are all mangled together. 

#### Example of success
Screenshot 2025-12-15 at 14 32 38

#### Example of failure
Screenshot 2025-12-15 at 14 31 42

#### Example of success

Screenshot 2025-12-15 at 14 23 21

#### Example of failure

Screenshot 2025-12-15 at 14 22 14

==COMMIT_MSG==
Wrap `@GradlePluginTests` TestKit build outputs in nice color-coded boxes
==COMMIT_MSG==

## Possible downsides?

There's a concern that this would  interfere with Intellij's diff formatter. The diff formatter only triggers on `assertEquals` ([source](https://stackoverflow.com/a/10945655)), and people are not asserting equality on entire build outputs. So it should be fine.

Also, ANSI codes will not be added to the output if the `CI` environment variable is set. This way we don't clobber CI with ANSI codes

[^1]: e.g. when testing for configuration cache compatibility, two builds are run, with the second time being a `--dry-run`

### DIFF
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
@@ -44,8 +44,7 @@ public final class DefaultGradleInvocation implements GradleInvocation {
             return new InvocationResult(result);
         } catch (UnexpectedBuildFailure e) {
             printFormattedOutput(false);
-            // Don't rethrow e, as it needlessly prints the inner build output again
-            throw new RuntimeException("Expected build to be successful: " + outputTitle);
+            throw e;
         }
     }
 
@@ -57,8 +56,7 @@ public final class DefaultGradleInvocation implements GradleInvocation {
             return new InvocationResult(result);
         } catch (UnexpectedBuildSuccess e) {
             printFormattedOutput(true);
-            // Don't rethrow e, as it needlessly prints the inner build output again
-            throw new RuntimeException("Expected build to fail: " + outputTitle);
+            throw e;
         }
     }
 
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
@@ -16,23 +16,58 @@
 
 package com.palantir.gradle.testing.execution;
 
+import java.io.ByteArrayOutputStream;
+import java.nio.charset.StandardCharsets;
+import org.gradle.testkit.runner.BuildResult;
 import org.gradle.testkit.runner.GradleRunner;
+import org.gradle.testkit.runner.UnexpectedBuildFailure;
+import org.gradle.testkit.runner.UnexpectedBuildSuccess;
 
 public final class DefaultGradleInvocation implements GradleInvocation {
 
     private final GradleRunner gradleRunner;
+    private final ByteArrayOutputStream capturedOutput;
+    private final String outputTitle;
 
-    public DefaultGradleInvocation(GradleRunner gradleRunner) {
+    public DefaultGradleInvocation(
+            GradleRunner gradleRunner, String outputTitle, ByteArrayOutputStream capturedOutput) {
         this.gradleRunner = gradleRunner;
+        this.outputTitle = outputTitle;
+        this.capturedOutput = capturedOutput;
     }
 
     @Override
     public InvocationResult buildsSuccessfully() {
-        return new InvocationResult(gradleRunner.build());
+        try {
+            BuildResult result = gradleRunner.build();
+            printFormattedOutput(true);
+            return new InvocationResult(result);
+        } catch (UnexpectedBuildFailure e) {
+            printFormattedOutput(false);
+            // Don't rethrow e, as it needlessly prints the inner build output again
+            throw new RuntimeException("Expected build to be successful: " + outputTitle);
+        }
     }
 
     @Override
     public InvocationResult buildsWithFailure() {
-        return new InvocationResult(gradleRunner.buildAndFail());
+        try {
+            BuildResult result = gradleRunner.buildAndFail();
+            printFormattedOutput(false);
+            return new InvocationResult(result);
+        } catch (UnexpectedBuildSuccess e) {
+            printFormattedOutput(true);
+            // Don't rethrow e, as it needlessly prints the inner build output again
+            throw new RuntimeException("Expected build to fail: " + outputTitle);
+        }
+    }
+
+    private void printFormattedOutput(boolean success) {
+        String output = capturedOutput.toString(StandardCharsets.UTF_8);
+        String formatted = success
+                ? OutputBoxFormatter.formatSuccess(outputTitle, output)
+                : OutputBoxFormatter.formatFailure(outputTitle, output);
+        // CHECKSTYLE.OFF: RegexpSinglelineJava
+        System.out.println("\n" + formatted + "\n");
     }
 }
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
@@ -60,12 +60,12 @@ public final class DefaultGradleInvocation implements GradleInvocation {
         }
     }
 
+    @SuppressWarnings("checkstyle:RegexpSinglelineJava")
     private void printFormattedOutput(boolean success) {
         String output = capturedOutput.toString(StandardCharsets.UTF_8);
         String formatted = success
                 ? OutputBoxFormatter.formatSuccess(outputTitle, output)
                 : OutputBoxFormatter.formatFailure(outputTitle, output);
-        // CHECKSTYLE.OFF: RegexpSinglelineJava
         System.out.println("\n" + formatted + "\n");
     }
 }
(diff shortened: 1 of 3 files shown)


felixdesouza:
this looks really cool!

> Also, I've verified manually that this does not cause CircleCI test failures to be misformatted. (see the failing caswe of PluginTestingJunitPluginTest [here](https://app.circleci.com/pipelines/github/palantir/gradle-plugin-testing/1431/workflows/64ce48b7-491a-4238-b9a4-9436b3428655/jobs/3109/tests))

however, this isn't true, have a look [here](https://output.circle-artifacts.com/output/job/4cee30aa-2dff-4093-83de-9f133f9d5012/artifacts/0/~/artifacts/junit/gradle-plugin-testing-junit/test/classes/com.palantir.gradle.testing.ete.ConfigurationCacheParameterTest.html)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/OutputBoxFormatter.java
```diff
@@ -0,0 +1,145 @@
...
+     * @param title the title to display in the header
+     * @param content the content to box
+     * @return the formatted output with red borders
+     */
+    public static String formatFailure(String title, String content) {
+        return format(title, content, ANSI_RED);
+    }
+
+    private static String format(String title, String content, String color) {
+        String[] lines = content.split("\\r?\\n");
```
felixdesouza: this is a little weird? there's a `lines` method on `String`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/OutputBoxFormatter.java
```diff
@@ -0,0 +1,145 @@
...
+
+        // Calculate the width based on content, but with reasonable bounds
+        int contentWidth = calculateWidth(title, lines);
+
+        StringBuilder result = new StringBuilder();
```
felixdesouza, line 70: regarding stripping ansi you can use something like: https://github.com/fusesource/jansi to do that for you. It's pretty neat, I've used it in an internal library.

What you'd end up doing is:

```java
AnsiConsole.out().println(...);
```

There's also some useful utility functions to help make writing/reading this stuff a bit easier. Here's a snippet our friend claude cooked up to convert:

```java
public static String bla(String args...) {
    String[] lines = content.split("\\r?\\n");

    // Calculate the width based on content, but with reasonable bounds
    int contentWidth = calculateWidth(title, lines);

    StringBuilder result = new StringBuilder();

    // Top border with title
    String header = " " + title + " ";
    result.append(ansi().fg(color)
            .a(TOP_LEFT)
            .a(header)
            .a(repeatChar(HORIZONTAL, contentWidth - header.length()))
            .a(TOP_RIGHT)
            .reset());
    result.append('\n');

    // Separator line
    result.append(ansi().fg(color)
            .a(LEFT_T)
            .a(repeatChar(HORIZONTAL, contentWidth))
            .a(RIGHT_T)
            .reset());
    result.append('\n');

    // Content lines - no truncation, pad to box width
    for (String line : lines) {
        result.append(ansi().fg(color).a(VERTICAL).reset());
        result.append(' '); // left padding

        result.append(ansi().fgBright(Color.WHITE).a(line).reset());
        // Pad to match box width
        result.append(repeatChar(' ', contentWidth - (2 * PADDING) - line.length()));

        result.append(' '); // right padding
        result.append(ansi().fg(color).a(VERTICAL).reset());
        result.append('\n');
    }

    // Bottom border
    result.append(ansi().fg(color)
            .a(BOTTOM_LEFT)
            .a(repeatChar(HORIZONTAL, contentWidth))
            .a(BOTTOM_RIGHT)
            .reset());

    return result.toString();
}

```

I'm sure you could probably make it look a bit nicer.

Not strictly required, but I think we should still use the `Ansi.out()` when we're actually emitting the code

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
```diff
@@ -16,23 +16,56 @@
...
+        }
+    }
+
+    private void printFormattedOutput(boolean success) {
+        String output = capturedOutput.toString(StandardCharsets.UTF_8);
+        String formatted = success
+                ? OutputBoxFormatter.formatSuccess(outputTitle, output)
+                : OutputBoxFormatter.formatFailure(outputTitle, output);
+        // CHECKSTYLE.OFF: RegexpSinglelineJava
+        System.out.println("\n" + formatted + "\n");
```
felixdesouza: You can replace this bit with:

```java
AnsiConsole.out().println("\n%s\n".formatted(formatted));
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/OutputBoxFormatter.java
kelvinou01: yeah, changed this to use `lines` instead

kelvinou01:
https://output.circle-artifacts.com/output/job/9c7e25eb-3e53-44da-9c1f-65657332f74a/artifacts/0/~/artifacts/junit/gradle-plugin-testing-junit/test/classes/com.palantir.gradle.testing.ete.ConfigurationCacheParameterTest.html

CI outputs are now working fine

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
kelvinou01: I think we discussed in person that unfortunately we couldn't include jansi into the program, as Gradle bundles an older version of jansi. 

I've replicated the functionality we need manually (adding ANSI color formatting, conditional of whether stdout is a terminal)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
```diff
@@ -16,23 +16,56 @@
...
+            throw e;
+        }
+    }
+
+    private void printFormattedOutput(boolean success) {
+        String output = capturedOutput.toString(StandardCharsets.UTF_8);
+        String formatted = success
+                ? OutputBoxFormatter.formatSuccess(outputTitle, output)
+                : OutputBoxFormatter.formatFailure(outputTitle, output);
+        // CHECKSTYLE.OFF: RegexpSinglelineJava
```
FinlayRJW: what is the advantage vs `SuppressWarnings`? Do we need todo a `// CHECKSTYLE.ON` afterwards?

kelvinou01: lets use SuppressWaarnings instead!

FinlayRJW:
👍

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -28,10 +33,14 @@ record DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) im
...
     public GradleInvocation with(Options options) {
+        ByteArrayOutputStream capturedOutput = new ByteArrayOutputStream();
+        Writer captureWriter = new OutputStreamWriter(capturedOutput, StandardCharsets.UTF_8);
+
         GradleRunner runner = GradleRunner.create()
                 .withProjectDir(rootProjectDir.toFile())
                 .withDebug(GradleInvoker.shouldRunInTestkitDebugMode())
-                .forwardOutput()
+                .forwardStdOutput(captureWriter)
+                .forwardStdError(captureWriter)
```
CRogers, line 43: Doesn't this have the effect of stopping the logs printing in real time? Previously I think they would go straight to stdout when the logs lines were produced, now I think it's going to wait until the invocation has ended before printing them out. I don't think this is too big a deal, but it's worth noting a behaviour change.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvocation.java
```diff
@@ -16,23 +16,56 @@
...
+            throw e;
+        }
     }
 
     @Override
     public InvocationResult buildsWithFailure() {
-        return new InvocationResult(gradleRunner.buildAndFail());
+        try {
+            BuildResult result = gradleRunner.buildAndFail();
+            printFormattedOutput(false);
```
CRogers, line 55: Is it going to be a little confusing that it's going to print a red box for an _expected_ failure? And also a green box for an _unexpected_ success (two lines below). Perhaps we need to call out the failure is expected (or success is unexpected) somehow? Or perhaps slightly different colours for these cases? I at least associate red/green with bad/good.

kelvinou01: > Is it going to be a little confusing that it's going to print a red box for an expected failure? And also a green box for an unexpected success (two lines below)

Yeah I thought about making the colour reflect the test expectation, but I think if we document/get used to red=failed build, green=successful build, it'll be fine?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
kelvinou01: > Doesn't this have the effect of stopping the logs printing in real time?

yep, unfortunately formatting logs in any nice way does remove the real time feedback. we should take note of this

PR #330, opened 2026-01-05
Add `@AddtionallyRunWithGradle` annotation for test-specific Gradle version testing

FinlayRJW:
## Before this PR

When writing tests for Gradle plugins, all test classes run against the same set of Gradle versions configured globally via `gradle/gradle-test-versions.yml` or the `gradleTestUtils` extension. There was no way to add extra Gradle versions for specific test classes that need to verify compatibility with additional versions beyond the global configuration.

## After this PR

Test classes and individual test methods can now use `@AddtionallyRunWithGradle` to specify extra Gradle versions to test against:
```java
@GradlePluginTests
@AddtionallyRunWithGradle{"7.6.5", "8.0"})
class CompatibilityTest {
    @Test
    void works_on_older_gradle_versions(GradleInvoker gradle, RootProject project) {
        // Runs against global versions + 7.6.5 + 8.0 (from class) 
    }

    @Test
    @AddtionallyRunWithGradle("8.5")
    void test_specific_version(GradleInvoker gradle, RootProject project) {
        // Runs against global versions + 7.6.5 + 8.0 (from class) + 8.5 (from method)
    }
}
```
Versions are merged with the global configuration and duplicates are automatically deduplicated.

When `@AddtionallyRunWithGradle` is applied at the method level, the additional versions specified are only run for that specific method. Other test methods in the same class will skip those method-specific versions via an `ExecutionCondition`.

This approach is necessary because JUnit's parameterized test template generates test invocations for all Gradle versions at the class level. Without the condition, every test method would run against every version from every method-level annotation in the class. The `AddtionallyRunWithGradleCondition`
filters out versions that don't belong to a particular method, ensuring method-level versions only apply to their intended test.

==COMMIT_MSG==
Add `@AddtionallyRunWithGradle` annotation

Adds a new annotation that allows test classes and individual test methods to
specify additional Gradle versions to test against, beyond those configured
globally. Versions are merged with the global configuration and deduplicated.
==COMMIT_MSG==

Possible downsides?

None anticipated.

### DIFF
--- a/docs/testing-guide.md
+++ b/docs/testing-guide.md
@@ -147,11 +147,14 @@ See [Resolution of Gradle versions to test against](../README.md#resolution-of-g
 
 #### Adding Versions for Specific Tests
 
-Use `@AdditionalGradleVersions` to add extra Gradle versions for a specific test class or individual test methods:
+Use `@WithGradleVersions` to add extra Gradle versions for a specific test class or individual test methods.
+
+> **Note:** This annotation is intended for exceptional cases where a specific test needs additional versions.
+> For configuring Gradle versions across your entire test suite, prefer setting versions in the `gradle/gradle-test-versions.yml` file.
 
 ```java
 @GradlePluginTests
-@AdditionalGradleVersions({"7.6.5", "8.0"})
+@WithGradleVersions({"7.6.5", "8.0"})
 class CompatibilityTest {
     @Test
     void works_on_older_gradle_versions(GradleInvoker gradle, RootProject project) {
@@ -159,14 +162,14 @@ class CompatibilityTest {
     }
 
     @Test
-    @AdditionalGradleVersions("8.5")
+    @WithGradleVersions("8.5")
     void test_specific_version(GradleInvoker gradle, RootProject project) {
         // This test runs against globally configured versions PLUS 7.6.5, 8.0 (from class), and 8.5 (from method)
     }
 }
 ```
 
-The versions from `@AdditionalGradleVersions` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
+The versions from `@WithGradleVersions` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
 
 ## File Operations
 
--- a/docs/testing-guide.md
+++ b/docs/testing-guide.md
@@ -147,14 +147,14 @@ See [Resolution of Gradle versions to test against](../README.md#resolution-of-g
 
 #### Adding Versions for Specific Tests
 
-Use `@WithSpecialCaseGradleVersions` to add extra Gradle versions for a specific test class or individual test methods.
+Use `@AdditionallyRunWithGradle` to add extra Gradle versions for a specific test class or individual test methods.
 
 > **Note:** This annotation is intended for exceptional cases where a specific test needs additional versions.
 > For configuring Gradle versions across your entire test suite, prefer setting versions in the `gradle/gradle-test-versions.yml` file.
 
 ```java
 @GradlePluginTests
-@WithSpecialCaseGradleVersions({"7.6.5", "8.0"})
+@AdditionallyRunWithGradle({"7.6.5", "8.0"})
 class CompatibilityTest {
     @Test
     void works_on_older_gradle_versions(GradleInvoker gradle, RootProject project) {
@@ -162,14 +162,14 @@ class CompatibilityTest {
     }
 
     @Test
-    @WithSpecialCaseGradleVersions("8.5")
+    @AdditionallyRunWithGradle("8.5")
     void test_specific_version(GradleInvoker gradle, RootProject project) {
         // This test runs against globally configured versions PLUS 7.6.5, 8.0 (from class), and 8.5 (from method)
     }
 }
 ```
 
-The versions from `@WithSpecialCaseGradleVersions` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
+The versions from `@AdditionallyRunWithGradle` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
 
 ## File Operations
 
--- a/docs/testing-guide.md
+++ b/docs/testing-guide.md
@@ -147,7 +147,7 @@ See [Resolution of Gradle versions to test against](../README.md#resolution-of-g
 
 #### Adding Versions for Specific Tests
 
-Use `@AdditionalGradleVersions` to add extra Gradle versions for a specific test class:
+Use `@AdditionalGradleVersions` to add extra Gradle versions for a specific test class or individual test methods:
 
 ```java
 @GradlePluginTests
@@ -157,10 +157,16 @@ class CompatibilityTest {
     void works_on_older_gradle_versions(GradleInvoker gradle, RootProject project) {
         // This test runs against the globally configured versions PLUS 7.6.5 and 8.0
     }
+
+    @Test
+    @AdditionalGradleVersions("8.5")
+    void test_specific_version(GradleInvoker gradle, RootProject project) {
+        // This test runs against globally configured versions PLUS 7.6.5, 8.0 (from class), and 8.5 (from method)
+    }
 }
 ```
 
-The versions from `@AdditionalGradleVersions` are merged with the globally configured versions. Duplicate versions are automatically deduplicated.
+The versions from `@AdditionalGradleVersions` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
 
 ## File Operations
 
--- a/docs/testing-guide.md
+++ b/docs/testing-guide.md
@@ -147,14 +147,14 @@ See [Resolution of Gradle versions to test against](../README.md#resolution-of-g
 
 #### Adding Versions for Specific Tests
 
-Use `@WithGradleVersions` to add extra Gradle versions for a specific test class or individual test methods.
+Use `@WithSpecialCaseGradleVersions` to add extra Gradle versions for a specific test class or individual test methods.
 
 > **Note:** This annotation is intended for exceptional cases where a specific test needs additional versions.
 > For configuring Gradle versions across your entire test suite, prefer setting versions in the `gradle/gradle-test-versions.yml` file.
 
 ```java
 @GradlePluginTests
-@WithGradleVersions({"7.6.5", "8.0"})
+@WithSpecialCaseGradleVersions({"7.6.5", "8.0"})
 class CompatibilityTest {
     @Test
     void works_on_older_gradle_versions(GradleInvoker gradle, RootProject project) {
@@ -162,14 +162,14 @@ class CompatibilityTest {
     }
 
     @Test
-    @WithGradleVersions("8.5")
+    @WithSpecialCaseGradleVersions("8.5")
     void test_specific_version(GradleInvoker gradle, RootProject project) {
         // This test runs against globally configured versions PLUS 7.6.5, 8.0 (from class), and 8.5 (from method)
     }
 }
 ```
 
-The versions from `@WithGradleVersions` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
+The versions from `@WithSpecialCaseGradleVersions` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
 
 ## File Operations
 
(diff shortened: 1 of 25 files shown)


docs/testing-guide.md
```diff
@@ -145,6 +145,23 @@ If not specified, tests run against the default version (currently `8.14.3`).
...
+```java
+@GradlePluginTests
+@AdditionalGradleVersions({"7.6.5", "8.0"})
+class CompatibilityTest {
+    @Test
+    void works_on_older_gradle_versions(GradleInvoker gradle, RootProject project) {
+        // This test runs against the globally configured versions PLUS 7.6.5 and 8.0
+    }
+}
+```
```
kelvinou01, line 170: Would be nice to have it work for specific test methods as well. Not sure how technically feasible this idea is though

FinlayRJW: good point will have a look

FinlayRJW: okay so I have updated so it can be applied to the method, but this creates a skipped version on all other tests, I couldn't think of another way todo this while still using `ClassTemplateInvocationContextProvider`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/AdditionalGradleVersionsCondition.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private Set<String> getVersionsForMethod(ExtensionContext context) {
+        Set<String> versions = new LinkedHashSet<>();
+
+        // Add base versions from configuration
+        context.getConfigurationParameter("com.palantir.gradle.testing.gradle_versions_to_test")
+                .ifPresent(param -> {
+                    List<String> configuredVersions = Splitter.on(',').splitToList(param);
+                    versions.addAll(configuredVersions);
+                });
```
felixdesouza: this is weird to do, you're spreading this logic across multiple places. At the very least, use a method the same way you've used `GradleVersionStore`.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/AdditionalGradleVersionsTest.java
```diff
@@ -0,0 +1,101 @@
...
+                .selectors(DiscoverySelectors.selectClass(AdditionalGradleVersionsFixtureTest.class))
+                // Base version configured via parameter
+                .configurationParameter("com.palantir.gradle.testing.gradle_versions_to_test", "7.6.5")
+                .configurationParameter("com.palantir.gradle.testing.configuration_cache_enabled", "false")
+                .execute();
+
+        List<Event> finished = executionResults.testEvents().finished().stream().toList();
+
+        // Should have 3 test runs: 7.6.5 from config + 8.0 and 8.5 from @AdditionalGradleVersions
+        assertThat(finished).hasSize(3);
```
felixdesouza: this is a little strange, let's just make the assertions more straightforward than do things in comments.

i.e.

```java
Assertions.assertThat(finished).satisfiesExactlyInAnyOrder(
  event -> Assertions.assertThatRanWithCorrectGradleVersion(
                AdditionalGradleVersionsFixtureTest.class, finished.get(0), "7.6.5"),
  ...)
```

Which is exactly what you want, there's going to be three, and there's not going to be more and you don't need to comment to explain math.

You can add the comments to the individual assertions and that's probably clearer as well.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/AdditionalGradleVersionsTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+        List<Event> finished = executionResults.testEvents().finished().stream().toList();
+
+        // Should have 2 test runs: 8.0 (deduplicated) and 8.5 from @AdditionalGradleVersions
+        assertThat(finished).hasSize(2);
+
+        Assertions.assertThatRanWithCorrectGradleVersion(
+                AdditionalGradleVersionsFixtureTest.class, finished.get(0), "8.0");
+        Assertions.assertThatRanWithCorrectGradleVersion(
+                AdditionalGradleVersionsFixtureTest.class, finished.get(1), "8.5");
```
felixdesouza: same thing here

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/AdditionalGradleVersionsTest.java
```diff
@@ -0,0 +1,101 @@
...
+        assertThat(skipped).hasSize(1);
+
+        // Verify the skipped test is the method without annotation running on 8.5
+        assertThat(skipped.get(0).getTestDescriptor().getDisplayName()).contains("test without method annotation");
+        assertThat(skipped.get(0)
+                        .getTestDescriptor()
+                        .getParent()
+                        .map(TestDescriptor::getDisplayName)
+                        .orElse(""))
+                .isEqualTo("Gradle 8.5");
```
felixdesouza: can do the same thing here, you just move the assertions in the satisfy block again

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/AdditionalGradleVersionsCondition.java
FinlayRJW: moved to a method in `GradleVersioningClassTemplate`

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/AdditionalGradleVersionsTest.java
FinlayRJW: good idea, have moved be more like

```java
        assertThat(finished)
                .satisfiesExactlyInAnyOrder(
                        // from config parameter
                        ranWithGradleVersion("7.6.5"),
                        // from @AdditionalGradleVersions
                        ranWithGradleVersion("8.0"),
                        // from @AdditionalGradleVersions
                        ranWithGradleVersion("8.5"));
```

docs/testing-guide.md
```diff
@@ -145,6 +145,32 @@ If not specified, tests run against the default version (currently `8.14.3`).
...
+#### Adding Versions for Specific Tests
+
+Use `@WithGradleVersions` to add extra Gradle versions for a specific test class or individual test methods.
+
+> **Note:** This annotation is intended for exceptional cases where a specific test needs additional versions.
+> For configuring Gradle versions across your entire test suite, prefer setting versions in the `gradle/gradle-test-versions.yml` file.
+
+```java
+@GradlePluginTests
+@WithGradleVersions({"7.6.5", "8.0"})
```
FinlayRJW: not sure this is the best name

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WithGradleVersions.java
```diff
@@ -0,0 +1,49 @@
...
+ * <p><b>Important:</b> This annotation is intended for exceptional cases where a specific test or test class
+ * needs to run against additional Gradle versions beyond the globally configured versions. For configuring
+ * Gradle versions across your entire test suite, prefer setting versions in the
+ * {@code gradle/gradle-test-versions.yml} file
+ *
+ * <p>The versions specified in this annotation will be merged with the versions configured in the yml file.
+ */
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface WithGradleVersions {
```
kelvinou01: > This annotation is intended for exceptional cases

I liked the idea of baking "temporary" into the name, to signal that there is a "more proper" way. Maybe `WithExceptionalGradleVersions` or something.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/AdditionallyRunWithGradleCondition.java
```diff
@@ -0,0 +1,52 @@
...
+ * For methods without the annotation, only base and class-level versions run.
+ */
+final class WithGradleVersionsCondition implements ExecutionCondition {
+
+    @Override
+    public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+        // Only evaluate for test methods
+        if (context.getTestMethod().isEmpty()) {
+            return ConditionEvaluationResult.enabled("No test method");
+        }
```
kelvinou01, line 39: I am guessing this is a noop?

kelvinou01 (commented):
Overall LGTM, leaving it unapproved so others can take a look

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WithGradleVersions.java
kelvinou01: On a separate note, maybe the name should indicate that this version is in addition to whatever base gradle versions you have configured.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/AdditionallyRunWithGradle.java
```diff
@@ -0,0 +1,49 @@
...
+     * The additional Gradle versions to test against.
+     * @return an array of Gradle version strings (e.g., "7.6.5", "8.0")
+     */
+    String[] value();
+
+    /**
+     * Optional reason explaining why these additional Gradle versions are needed.
+     * @return the reason for requiring these specific versions
+     */
+    String reason() default "";
```
kelvinou01, line 48: To make our lives easier in the future, maybe reason should be required?

kelvinou01: Though I can see how this could be annoying

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/AdditionallyRunWithGradleCondition.java
FinlayRJW: yeah

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/AdditionallyRunWithGradle.java
FinlayRJW: I'm 50/50 on this as I feel 90% of the reasons will just be "this test needs to run against version X of gradle" which doesn't really give much signal

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WithGradleVersions.java
FinlayRJW: naming is so hard what about `WithSpecialCaseGradleVersions`? I think that implies it should not be generally used? Not sure it quite hits the additive nature but `WithAdditionalSpecialCaseGradleVersions` felt a bit wordy!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/AdditionallyRunWithGradle.java
kelvinou01: > 90% of the reasons will just be "this test needs to run against version X of gradle" which doesn't really give much signal

So real 😭

kelvinou01: Alright then I think leaving it optional will do

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WithGradleVersions.java
kelvinou01: I think as long as it probes people to think "hmm, is this not the canonical/default way?" it should be fine!

Maybe `@AddtionallyRunWithGradle({"8.8", "8.14.3"})`?

docs/testing-guide.md
```diff
@@ -145,6 +145,32 @@ If not specified, tests run against the default version (currently `8.14.3`).
 
 See [Resolution of Gradle versions to test against](../README.md#resolution-of-gradle-versions-to-test-against) for more details.
 
+#### Adding Versions for Specific Tests
+
+Use `@AdditionallyRunWithGradle` to add extra Gradle versions for a specific test class or individual test methods.
+
+> **Note:** This annotation is intended for exceptional cases where a specific test needs additional versions.
+> For configuring Gradle versions across your entire test suite, prefer setting versions in the `gradle/gradle-test-versions.yml` file.
+
```
kelvinou01, line 154: Nice!

kelvinou01 (approved):

PR #332, opened 2026-01-05
Read gradle/gradle-test-versions.yml from the root dir, rather than project dir

kelvinou01:
## Before this PR

gradle/gradle-test-versions.yml was being read from child project directories, rather than the root directory. This meant it couldn't be found. 

## After this PR

We had a [test](https://github.com/palantir/gradle-plugin-testing/blob/develop/gradle-plugin-testing/src/test/groovy/com/palantir/gradle/plugintesting/PluginTestingPluginIntegrationSpec.groovy#L165) for this. IntegrationSpec uses single-project builds, which means the settings and project dirs are the same. Hence the test passed.

However, in a multi-project build, we have to read from the settings directory instead.  

==COMMIT_MSG==
Read gradle/gradle-test-versions.yml from the settings dir, rather than project dir
==COMMIT_MSG==


FinlayRJW:
👍

PR #334, opened 2026-01-05
Add `@WithJdkAutomanagement` to junit-testing

crogoz:
## Before this PR

Not an easy way to run the tests with a specific version of jdks.

## After this PR

Adds `@WithJdkAutomanagement` annotation that can be applied to GradlePluginTests to set up the jdks required to run the daemonTarget, compilation/runtime etc. tasks. 

The annotation: 
- configures the JDK automanagement (see more: [here](https://github.com/palantir/gradle-jdks/blob/develop/gradle-jdks-setup/README.md)); adds `jdks` & `jdks.settings` plugins; sets `palantir.jdk.setup.enabled=true`
- runs `setupJdks` that generates & sets up all the jdks:
Screenshot 2026-01-12 at 11 53 45 AM

FLUP for https://github.com/palantir/gradle-baseline/pull/3379#issue-3672523716 (started the move [here](https://github.com/palantir/gradle-baseline/compare/cr/testing-plugin?expand=1)). Will also be used to test gradle-jdks.

==COMMIT_MSG==
Add `@WithJdkAutomanagement` to junit-testing
==COMMIT_MSG==


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleWithJdksInvocation.java
```diff
@@ -0,0 +1,106 @@
...
+    public void setupJdkAutomanagement() {
+        try {
+            setupInvocation.buildsSuccessfully();
+            runWithLogger(generateToolchainsInvocation);
+        } catch (Exception e) {
+            throw new GradleWithJdksInvocationFailure(e);
+        }
+    }
+
+    public static void runWithLogger(ProcessBuilder processBuilder) {
```
crogoz: same as CommandRunner.runWithLogger (in gradle-jdks). If a jdk needs to be installed, the ouput lines will be shown in the terminal. 

Image

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,112 @@
...
+    public GradleInvocation withArgs(String... args) {
+        setupRootProject(rootProject);
+
+        GradleInvocation generateGradleJdkConfigs = gradleInvoker.withArgs("generateGradleJdkConfigs");
+
+        ProcessBuilder processBuilder = new ProcessBuilder()
+                .command("./gradle/gradle-jdks-setup.sh")
+                .directory(rootProject.path().toFile());
+        return new GradleWithJdksInvocation(
+                generateGradleJdkConfigs, processBuilder, () -> getInvokerWithToolchainsConfigured(args));
```
crogoz: `() -> getInvokerWithToolchainsConfigured(args))`

this is a supplier because I don't know at this time what the value/path for the gradle java home is.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    // this needs to be called after the daemon jdk version is rendered (by generateGradleJdkConfigs)
+    private GradleInvocation getInvokerWithToolchainsConfigured(String... args) {
+        String[] withJavaHome = ImmutableList.<String>builder()
+                .add(args)
+                .add(String.format("-Dorg.gradle.java.home=%s", getGradleJavaHome(rootProject.path())))
```
crogoz, line 63: Opted for using a gradle runner instead of the `./gradlew` process as we would be able to debug the gradle runner. 
In order to set the correct Gradle java home, I am using the property `org.gradle.java.home`

crogoz:
depends on https://github.com/palantir/gradle-jdks/pull/653

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -27,7 +27,18 @@ public interface GradleInvoker {
 
     GradleInvocation withArgs(String... args);
 
-    static GradleInvoker create(Path path, GradleVersion gradleVersion, boolean configurationCache) {
+    GradleVersion getGradleVersion();
```
CRogers, line 30: Can we just call this `gradleVersion()`? Not a big fan of the `get` prefix, don't think it adds much. And we avoid having `gradleVersion()` and `getGradleVersion()` on `DefaultGradleInvoker`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleWithJdksInvocation.java
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.testing.execution;
+
+import java.util.concurrent.Callable;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
+
+public record GradleWithJdksInvocation(GradleInvocation setupInvocation, Callable<GradleInvocation> tasksInvocation)
```
CRogers, line 23: Can this be package-private?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Override
+    public GradleVersion getGradleVersion() {
+        return gradleInvoker.getGradleVersion();
+    }
+
+    @Override
+    public GradleInvocation withArgs(String... args) {
+        setupRootProject(rootProject);
```
CRogers, line 53: This means we add the properties/plugin applications every single time we run Gradle. Is this safe? Does it add the property/plugins multiple times?

I'm wondering whether this should be something we do once at test initialisation rather than every single gradle invocation?

CRogers: Like in the `beforeEach` of `WithJdkAutomanagementExtension`?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleWithJdksInvoker.java
```diff
@@ -0,0 +1,112 @@
...
+                        .orElseGet(() -> System.getProperty("user.home") + "/.gradle"))
+                .resolve("gradle-jdks")
+                .resolve(localJdkPath);
+    }
+
+    private static void setupRootProject(RootProject rootProject) {
+        rootProject.gradlePropertiesFile().appendProperty("palantir.jdk.setup.enabled", "true");
+        rootProject.buildGradle().plugins().add("java");
+        rootProject.buildGradle().plugins().add("com.palantir.jdks");
+        rootProject.settingsGradle().plugins().add("com.palantir.jdks.settings");
```
CRogers, line 110: I have a feeling that having this logic coupled inside of gradle-plugin-testing is going to bite us in the future. For example:

* To support Gradle 9, we're probably going to need to set `org.gradle.java.installations.auto-detect=false` and `org.gradle.java.installations.auto-download=false` in the gradle properties
* If we do the break and remove the old gradle-jdks workflow that uses baseline-java-versions to set the toolchains everywhere, we won't need the `palantir.jdk.setup.enabled`.
* Maybe we need to change how we setup jdks before running Gradle etc? Perhaps we end up using the built in Gradle tooling eventually?

The ideal situation would be to have all the logic for this extension (or at least some of the logic) in the gradle-jdks repo. This means the version of `@WithJdkAutomanagement` junit extension people use would be matched up with the version of gradle-jdks they are testing against. They would add a `testImplementation 'com.palantir.gradle.jdks:gradle-plugin-testing-jdks'` and then use the `@WithJdkAutomanagement` extension from in there.

It's probably worth having a go or at least having a think about how we could have this extension in the `gradle-jdks` repo and what the sensible API for this would be. We already have a similar extension (the configuration cache extension) that changes how gradle is invoked - perhaps there's a more generic extension hook we can write to enable both gradle-jdks support and configuration cache? Like some kind of hook or extension point that changes how Gradle is invoked? That way we might be able to untangle all the GradleInvoker subclasses we need for configuration cache + jdk automanagement.

Another thought: I think with the logic in `gradle-jdks` too we could also automatically mark every test that uses `@WithJdkAutomanagement` as `@DisabledConfigurationCache`, then when we do make `gradle-jdks` configuration cache compatible we can just remove that marking in the testing extension, and we don't need to manually remove it on lots of tests.

crogoz:
closing in favor of https://github.com/palantir/gradle-plugin-testing/pull/344

PR #337, opened 2026-01-06
Add `@RestrictToGradleVersionsEqualTo` annotation for filtering test matrix

FinlayRJW:
## Before this PR

When using `@RestrictToGradleVersionsEqualTo`, you can only add versions to the test matrix. There's no way to restrict a specific test to run against only certain versions from the matrix. If you need to run a test against just one Gradle version (e.g., for version-specific behavior testing), you had to run it against
all configured versions.

## After this PR

Introduces `@RestrictToGradleVersionsEqualTo` annotation that filters the test matrix to only include specified versions:

- Can be applied at class level (filters the entire test class matrix upfront) or method level (skips non-matching versions)
- Versions not in the matrix are simply not run (no error)
- Can be combined with `@RestrictToGradleVersionsEqualTo` to first add a version then filter to only that version: `@WithSpecialCaseGradleVersionsCondition("8.5")` `@RestrictToGradleVersionsEqualTo("8.5")`
- Supports an optional reason parameter to document why specific versions are required

==COMMIT_MSG==
Add `@RestrictToGradleVersionsEqualTo` annotation
==COMMIT_MSG==

## Possible downsides?

- If a user specifies a version in `@RestrictToGradleVersionsEqualTo` that isn't in the matrix, the test silently doesn't run (by design, but could be surprising)


docs/testing-guide.md
```diff
@@ -171,6 +171,44 @@ class CompatibilityTest {
 
 The versions from `@WithGradleVersions` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
 
+#### Filtering to Specific Versions
+
+Use `@WithOnlyGradleVersions` to filter the test matrix to only run on specific Gradle versions. Unlike `@WithGradleVersions` which adds versions, this annotation restricts which versions from the matrix will actually run.
```
kelvinou01: The name `@WithOnlyGradleVersions` could imply an authoritative list of versions you run the test with. This can be confusing.

FinlayRJW: what about `RestrictToGradleVersionsEqualTo` I think that implies it only filters?

FinlayRJW: then we can add `RestrictToGradleVersionsGreaterThan` etc later

PR #344, opened 2026-01-20
Allow registration of GradleInvokers using meta-annotations

crogoz:
## Before this PR

No support for extra GradleInvokers (eg. running with JDK automanagement enabled). work started in https://github.com/palantir/gradle-plugin-testing/pull/334, however the PR introduced tight coupling between the gradle-jdks implementation and gradle-plugin-testing.  (more discussion about this [here](https://github.com/palantir/gradle-plugin-testing/pull/334#discussion_r2698925963))

## After this PR

- Introduces a `GradleInvokerDecorator` that is used to wrap a `GradleInvoker` (using a `DecorationContext`) that adds extra/a specialized behaviour. Moved the configuration cache workflow to a new `ConfigurationCacheDecorator` registered when the `configuration cache` flag is enabled. 

- Introduces a meta-annotation `RegistersGradleInvokerDecorator` that can be applied to junit extensions (eg. `@WithJdkAutomanagement` in gradle-jdks - [WIP here](https://github.com/palantir/gradle-jdks/compare/cr/jdks-testing-module?expand=1)); these annotations will be automatically discovered and the corresponding Decorators will be used to register & apply the specialized GradleInvokers.

Users can create custom decorators for specialized Gradle testing invocations like so:  
```java
@target({ElementType.TYPE, ElementType.METHOD})
@retention(RetentionPolicy.RUNTIME)
@RegistersGradleInvokerDecorator(MySpecializedDecorator.class)
public @interface MySpecializedAnnotation {}
``` 

==COMMIT_MSG==
Allow registration of GradleInvokers using meta-annotations
==COMMIT_MSG==


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -38,6 +39,8 @@ public GradleInvocation withArgs(String... args) {
                 .withArguments(ImmutableList.<String>builder()
                         .addAll(Arrays.asList(args))
                         .add("--stacktrace")
+                        .add("-P__TESTING=true")
+                        .add("-P__TESTING_palantir.gradle.plugin.tests=true")
```
crogoz: I need to see whether [I am running inside the tests ](https://github.com/palantir/gradle-jdks/compare/develop...cr/jdks-testing-module#diff-874bce15e4b615599babbfed22b0a6b328b3e65dcfaef2acf53a28a29f07e7f5R74) - that workflow (running ./gradlew in a new process) is not compatible with the new `plugins {}` block.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,29 +16,42 @@
...
 import java.nio.file.Path;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import java.util.List;
+import org.junit.jupiter.api.extension.ExtensionContext;
 
 public interface GradleInvoker {
 
-    Logger log = LoggerFactory.getLogger(GradleInvoker.class);
+    GradleVersion gradleVersion();
```
crogoz: based on the gradleVersion we might have different error messages in the tests, so we need to know what gradleversion we ran on.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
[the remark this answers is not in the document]
crogoz: this is the setup of gradle-utils EnvVariables

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvoker.java
```diff
@@ -25,15 +25,14 @@
...
 
     @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
-    public ConfigurationCacheInvoker(Path rootProjectDir, GradleInvoker gradleInvoker) {
-        this.rootProjectDir = rootProjectDir;
-        this.gradleInvoker = gradleInvoker;
+    public ConfigurationCacheInvoker {}
+
+    @Override
+    public GradleVersion gradleVersion() {
+        return gradleInvoker.gradleVersion();
```
CRogers: Do we need `gradleVersion()` here? We never seem to use it here or in the gradle-jdks change? I think ideally we'd get the gradle version info in some other way (from an Extension store of some kind for other extensions, or maybe injected in as a method parameter if people need it for tests).

crogoz: I was thinking we might need it once we migrate the gradle-jdks tests (eg. [here](https://github.com/palantir/gradle-jdks/compare/cr/migration#diff-7c21971dfd2a34ea4067c9ca99fa091a22a65adf77d3e6c825f41d5c954bb2b3R232) - the message differs depending on the gradle version. atm we are only testing against 8.14)

I can remove it from this PR and introduce it in another one then.

CRogers: I think there's probably a better way than bubbling through all the invokers at least.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+    @SuppressWarnings("unchecked")
+    private static List<GradleInvokerDecorator> getOrCreateDecoratorList(ExtensionContext context) {
+        return (List<GradleInvokerDecorator>)
+                context.getStore(NAMESPACE).getOrComputeIfAbsent(DECORATORS_KEY, _key -> new ArrayList<>());
+    }
+
+    @SuppressWarnings("unchecked")
+    private static List<GradleInvokerDecorator> getDecoratorList(ExtensionContext context) {
+        return (List<GradleInvokerDecorator>) context.getStore(NAMESPACE).get(DECORATORS_KEY);
+    }
```
CRogers: Not really how (or if) this works when there are multiple `ExtensionContext`s? Like this:

```java
@DecoratorA
class TestClass {
    @DecoratorB
    void test_method() {
```

Will the `TestClass` and `test_method` have different extension contexts so have different lists of decorators? Ideally we'd have `@DecoratorA` in the `TestClass` context and `DecoratorB(DecoratorA(...))` in the `test_method` context?

CRogers: Feels like we need a test for multiple decorators at least - I'm also not 100% sure what the order of decoration should be/makes sense here.

CRogers: Also, what happens if we have the same decorator multiple times? eg

```java
@WithJdkAutomanagement
class TestClass {
    @WithJdkAutomanagement
    void test_method() {
```

...should this be an error? Should be only instantiate one?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+        getOrCreateDecoratorList(context).add(decorator);
+    }
+
+    /**
+     * Returns all registered decorators for the given extension context, in registration order.
+     *
+     * @param context the JUnit extension context
+     * @return an unmodifiable list of decorators in registration order
+     */
+    public static List<GradleInvokerDecorator> getDecorators(ExtensionContext context) {
```
CRogers: I think this `getDecorators` method is a little odd, as it just returns a list and then expects the caller to do the right thing. It might be nicer if there's just a `GradleInvoker decorate(ExtensionContext context, GradleInvoker gradle)` that will decorate it, perhaps traversing the extension contexts if it needs to.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvoker.java
CRogers: Like maybe it could be in the `DecoratorContext` if we needed to be.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorRegistry.java
crogoz: is this what you mean ? https://github.com/palantir/gradle-plugin-testing/pull/344/commits/123a702d7424c7da158cee8be91a4969d9fdbe72. I think the extensionContext will be different for the test and the classes (https://github.com/junit-team/junit-framework/blob/main/junit-jupiter-api/src/main/java/org/junit/jupiter/api/extension/ExtensionContext.java#L109-L115)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
CRogers: Because this is a gradle-jdks related thing - should we have the jdk automanagement gradle invoker set this instead?

CRogers: Maybe it just could detect whether `__TESTING` is set instead of having to set another property? Perhpas this could be a method on the `environment-variables` in gradle-utils.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -38,6 +39,8 @@ public GradleInvocation withArgs(String... args) {
                 .withArguments(ImmutableList.<String>builder()
                         .addAll(Arrays.asList(args))
                         .add("--stacktrace")
+                        .add("-P__TESTING=true")
```
CRogers, line 40: This is interesting, setting it by default here. It would at least ensure in all the plugins that use are in testing mode when we run their code. It probably makes sense to do this here. Might be worth adding the changelog.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorRegistry.java
crogoz: > Also, what happens if we have the same decorator multiple times? eg

edit: ~~Interesting, just tested and it seems like only the first decorator (the class decorator) is taken into consideration, the second one is a noop~~ Tested with the wrong setup. It does apply it multiple times. I changed it such that we throw when we get duplicate annotations both in classes/nested classes & the methods (https://github.com/palantir/gradle-plugin-testing/pull/344/commits/bc3220e86f953817b4a629e837eae2d647acb82f)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
crogoz: I think this flag would make sense for cases when people don't run with the `JdkAutomanagement` as well. Any time we run `setupJdks` inside the GradlePluginTests, we shouldn't run `./gradlew`. And this might be the case for other future changes. 

I could keep just `-P__TESTING` though for this and change the logic to only check for this property [here](https://github.com/palantir/gradle-jdks/pull/661/changes#diff-d14af7703cef8b606999ab3937204e21f209c24a65a6196c2c63b24ac0fd8c40R189)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorRegistry.java
crogoz: should be addressed in https://github.com/palantir/gradle-plugin-testing/pull/344/commits/bc3220e86f953817b4a629e837eae2d647acb82f

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
crogoz: > Perhpas this could be a method on the environment-variables in gradle-utils.

opened https://github.com/palantir/gradle-utils/pull/555 for this

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,29 +16,33 @@
...
 import java.nio.file.Path;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.junit.jupiter.api.extension.ExtensionContext;
 
 public interface GradleInvoker {
 
-    Logger log = LoggerFactory.getLogger(GradleInvoker.class);
-
     GradleInvocation withArgs(String... args);
```
CRogers: I'm a little concerned about decorating `GradleInvoker` now I come to think of it. If we ever wanted to add another method to `GradleInvoker`, like `withTestingEnvironmentVariable` or something similar, it would be a breaking for `gradle-jdks` who would have to implement and perhaps duplicate some logic...

CRogers: Actually, I guess adding other `with` methods here would be bad design. Instead, we should really have a single method that is passed a builder of some kind:

```java
gradle.with(Options.builder()
    .args(...)
    .testingEnvironmentVariable(...)
    .somethingElse(...)
    .builder()).buildsSuccessfully()
```

...that is just passed through all the delegates. To add this would be a breaking change, but I think only once at least, which we can manage.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/DiscoveredDecoratorsStore.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.lang.annotation.Annotation;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+
+/**
+ * Stores the mapping between test classes and their associated decorator annotations in JUnit extension context.
```
CRogers: What about if the decorator annotation is on a test method?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
```diff
@@ -0,0 +1,144 @@
...
+public final class GradleInvokerDecoratorDiscoveryExtension implements BeforeAllCallback, BeforeEachCallback {
+
+    private static final Logger log = LoggerFactory.getLogger(GradleInvokerDecoratorDiscoveryExtension.class);
+
+    @Override
+    public void beforeAll(ExtensionContext context) {
+        Class<?> currentClass = context.getRequiredTestClass();
+
+        Set<Class<? extends Annotation>> currentDiscoveredDecorators = discoverDecorators(currentClass);
+        Set<Class<? extends Annotation>> parentDecorators = Optional.ofNullable(currentClass.getEnclosingClass())
```
CRogers: Rather than going up enclosing classes, would it make more sense to use `context.getParent()` on the `ExtensionContext`? And make sure we just store the decorators in the `ExtensionContext` they are applied on? I think it may handle test methods out of the box then.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/DiscoveredDecoratorsStore.java
crogoz: I tested this [here](https://github.com/palantir/gradle-plugin-testing/pull/344/changes#diff-c4810ec357d66875a571e990673064afcd59cd0c072e297e367b8faeda2f0291R37). Is this testcase what you are referring to?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
CRogers: Ok, I've been reading this all wrong. I was hoping there was a way for us to just put something in the relevant `ExtensionContext`'s `Store` when the decorator annotation is placed on a test class/method etc. Then we could just traverse up the extension context stores to find all the decorators. But that first part doesn't seem possible - we need to use `beforeAll` (so that the data is in the in the Store if Gradle is invoked in a beforeAll?) and `beforeEach` for the test time.

But it all feels very brittle - it feels like we have one thing finding these and putting them in stores _hopefully just in time for_ the `GradleInvokerParameterResolver` to create it.

I wonder if instead, we don't have this `GradleInvokerDecoratorDiscoveryExtension` at all, and just do this discovery lazily when `GradleInvokerDecoratorRegistry#decorate` is called? That way we may not need any state and avoid possible timing issues (which are going to be really hard to debug)?

crogoz: I've simplified the logic [here](https://github.com/palantir/gradle-plugin-testing/pull/344/commits/5cb22b034b336f49284890c42454955b920a8b80) to use the context.getParent() and store the annotations per class level (aka extension level) so I don't need to keep a map. 

> And make sure we just store the decorators in the ExtensionContext they are applied on? I think it may handle test methods out of the box then.

I am not really sure if this implementation is what you had in mind. The test methods are handled already ootb (see comment https://github.com/palantir/gradle-plugin-testing/pull/344#discussion_r2717103053)

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/DuplicatedAnnotationsFailTest.java
```diff
@@ -0,0 +1,78 @@
...
+                .execute();
+
+        List<Event> testEvents = executionResults.testEvents().failed().list();
+        assertThat(testEvents).hasSize(2);
+
+        // Test 1: Method-level duplicate on duplicate_decorator_throws_exception
+        Event failedTest1 = testEvents.get(0);
+        assertExceptionMessage(
+                failedTest1,
+                "Decorator annotation @WithDecorator is already registered in a parent class. The same decorator"
```
CRogers: Nice

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
```diff
@@ -0,0 +1,91 @@
...
+                    println "Hello from task"
+                }
+            }
+            """);
+
+        // The decorator adds --info flag, so we should see verbose output
+        InvocationResult result = invoker.withArgs("hello").buildsSuccessfully();
+        result.assertThat().output().contains("Hello from task");
+        // With --info, we see "Task :hello" status line
+        result.assertThat().output().contains("Task :hello");
```
CRogers: Don't you see this without `--info`?

CRogers: Test passes without the annotation

CRogers: Same in the other tests

crogoz: addressed [here](https://github.com/palantir/gradle-plugin-testing/pull/344/changes/54747e0ae376207824ed31b47855c989f5049424)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
CRogers: Yeah, I don't think we can do with what I was intending by the part you quoted.

But the bit we might be able to do is avoid putting the decorators into stores at all. Couldn't we lazily discover the decorator annotations when we need to, ie when we need to make the `GradleInvoker`? This would prevent all possible bugs where the `GradleInvokerDecoratorDiscoveryExtension` doesn't store the decorations in time before the `GradleInvoker` is created, which I'm not sure will always happen. Also just removes a load of state.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/DuplicatedAnnotationsFailTest.java
```diff
@@ -0,0 +1,78 @@
...
+                        + " annotation cannot be applied at both class and method level.",
+                "Please remove it from the test method `duplicate_decorator_throws_exception`");
+
+        // Test 2: Method-level duplicate on another_test_with_duplicated_decorator
```
felixdesouza: just have separate tests?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
```diff
@@ -0,0 +1,145 @@
...
+        return context.getParent()
+                .map(enclosingClass -> DiscoveredDecoratorsStore.getStoredDecorators(context))
+                .orElseGet(Set::of);
+    }
+
+    /**
+     * Discovers decorator annotation types from an annotated element without registering them.
+     */
+    private Set<Class<? extends Annotation>> discoverDecorators(AnnotatedElement element) {
+        return Stream.of(element.getAnnotations())
```
felixdesouza: we should probably use `AnnotationSupport` to get annotations, I think it supports meta-annotations and also inherited annotations properly.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
```diff
@@ -0,0 +1,145 @@
...
+                .filter(parentDecorators::contains)
+                .findFirst()
+                .ifPresent(decorator -> {
+                    throw new IllegalStateException(String.format(
+                            "Decorator annotation @%s is already registered in a parent class. %s",
+                            decorator.getSimpleName(), message));
+                });
+    }
+
+    private static Set<Class<? extends Annotation>> getParentDecorators(ExtensionContext context) {
```
felixdesouza: don't you need to recursively get all the parents? we'd have to either guarantee that we're storing all the decorators from all parents inside the parent decorator context before `getParentDecorators` is called?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private static void throwForDuplicatedAnnotations(
+            ExtensionContext context, Set<Class<? extends Annotation>> currentDiscoveredDecorators, String message) {
+        Set<Class<? extends Annotation>> parentDecorators = getParentDecorators(context);
+        currentDiscoveredDecorators.stream()
+                .filter(parentDecorators::contains)
+                .findFirst()
+                .ifPresent(decorator -> {
+                    throw new IllegalStateException(String.format(
+                            "Decorator annotation @%s is already registered in a parent class. %s",
```
felixdesouza: we should probably show what class has already registered this decorator for easier debugging

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
```diff
@@ -0,0 +1,145 @@
...
+    @Override
+    public void beforeAll(ExtensionContext context) {
+        Class<?> currentClass = context.getRequiredTestClass();
+
+        Set<Class<? extends Annotation>> currentDiscoveredDecorators = discoverDecorators(currentClass);
+
+        throwForDuplicatedAnnotations(
+                context, currentDiscoveredDecorators, String.format("""
+                    The same decorator annotation cannot be applied at multiple class levels. Please remove the extra annotation from the class `%s`.
+                    """, currentClass.getSimpleName()));
```
felixdesouza: similar comment, it would be good to know where it's applied in parent contexts, suddenly you could end up breaking the test by removing it in the parent place.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
```diff
@@ -0,0 +1,145 @@
...
+                            registeringDecorator.value().getName(),
+                            annotation.annotationType().getSimpleName()),
+                    e);
+        }
+    }
+
+    private static void throwForDuplicatedAnnotations(
+            ExtensionContext context, Set<Class<? extends Annotation>> currentDiscoveredDecorators, String message) {
+        Set<Class<? extends Annotation>> parentDecorators = getParentDecorators(context);
+        currentDiscoveredDecorators.stream()
```
felixdesouza: I don't think it's enough to just consider the classes? You also have the concept of annotations having parameters, and that feeds into the decorator factory meaning you could have different implementations of the decorator based on that.

For the jdks example, that doesn't make any sense of course.

If you don't have the decorator factory, then you either probably don't need the `GradleInvokerDecoratorFactory` type at all? You can then just have `@RegistersGradleInvokerDecorator(<DecoratorType>)` instead of it pointing to the `GradleInvokerDecoratorFactory`.

I'm not really opposed to either, but it does make the semantics a little weird when you can have different instantiations of an annotation

felixdesouza: Imagine if you had a generic decorator which added some testing parameters, and you wanted to just stack them. Individually, they're separate, but you can stack if you wanted to!

e.g. 

```java
@GradlePluginTests
@WithTestingParameter("__TESTING=TRUE")
class MyTestClass {
  
  @Test
  @WithTestingParameter("SOME_TESTING_PARAMETER=BLA")
  @WithTestingParameter("ANOTHER_TESTING_PARAMETER=BLA")
  void myTest() {
    // do something
  }
}
```

How the api looks now, it seems to suggest that you will get a separate decorator for each instance of the `@WithTestingParameter` annotation, and in fairness they're separate.

A different angle is actually just treating them like `@Repeatable` annotations and you instead look at the container. But that does nothing for cross context annotations if I've understood correctly.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
```diff
@@ -0,0 +1,87 @@
...
+            """);
+
+        InvocationResult result = invoker.withArgs("hello").buildsSuccessfully();
+        result.assertThat().output().contains("Hello from task");
+        result.assertThat().output().contains("Task :help");
+    }
+
+    @Test
+    @WithArgAddingDecorator2(arg = "help")
+    @WithArgAddingDecorator3(arg = "-Pname=hello")
```
felixdesouza: yeah, I don't believe you should need to have 3 separate instances of `@WithArgAddingDecorator` and then to boot 3 more decorator factories, to get to the underlying `ArgAddingDecorator`. Perhaps you can have a flag that says that they don't stack across contexts, or just use the presence of the `@Repeatable` to infer that and you collect everything all together idk.

felixdesouza: another thing you could do, if you go down the repeatable approach is to then collect all the annotation instances, and then pass all the vlaues of it to the factory so then it combines it accordingly. You could have a default naive impl that just creates them in series.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
crogoz: I've updated the logic [here](https://github.com/palantir/gradle-plugin-testing/pull/344/commits/5c5bfa378e2d665d3b56e96940b0a1fd0097814d)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -27,18 +40,91 @@ public interface GradleInvoker {
...
+        return Stream.concat(inheritedAnnotations.stream(), extraAnnotations.stream())
+                .toList();
+    }
+
+    /**
+     * Collects all unique annotations starting from the current extension context (test method)
+     * and going up the parent tree (test class, parent classes, etc.).
+     */
+    private static Set<Annotation> collectAnnotationsFromContext(ExtensionContext context) {
+        return Stream.iterate(context, Objects::nonNull, ctx -> ctx.getParent().orElse(null))
```
crogoz: note: the hierarchy for contexts is `TestMethodExtensionContext` -> `ClassTemplateInvocationExtensionContext `-> `ClassExtensionContext`. 
Both `ClassTemplateInvocationExtensionContext` & `ClassExtensionContext` will have the same annotations.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -27,18 +40,91 @@ public interface GradleInvoker {
...
+                .flatMap(ctx -> Stream.concat(
+                        ctx.getTestMethod()
+                                .map(method -> getAnnotations(method.getAnnotations()))
+                                .orElseGet(Stream::empty),
+                        ctx.getTestClass()
+                                .map(clazz -> getAnnotations(clazz.getAnnotations()))
+                                .orElseGet(Stream::empty)))
+                .collect(Collectors.toSet());
```
crogoz: I am doing a de-duplication of the Annotations now (previously I was throwing if both a class & a method has the same annotation). Added tests for duplicated annotations [here](https://github.com/palantir/gradle-plugin-testing/pull/344/changes#diff-5f91bb14a2f754ce297d04011ca7f89ab15a4e790d5be30f363ccb8a07ce1f0dR100)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -27,18 +40,91 @@ public interface GradleInvoker {
...
+        }
+        return baseInvoker;
+    }
+
+    private static List<GradleInvokerDecorator> discoverDecorators(ExtensionContext extensionContext) {
+        List<GradleInvokerDecorator> inheritedAnnotations = collectAnnotationsFromContext(extensionContext).stream()
+                .map(GradleInvoker::getDecoratorFromAnnotation)
+                .toList();
+        List<GradleInvokerDecorator> extraAnnotations =
+                ConfigurationCacheStore.isConfigurationCacheEnabled(extensionContext)
```
crogoz: I am not sure if there is a better way for CC to be enabled

felixdesouza (commented):
had a couple of comments I made yesterday, not sure how many will be in date, but food for thought

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecorator.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public interface GradleInvokerDecorator {
+
+    /**
+     * Decorates the given invoker.
+     *
+     * @param context provides access to the root project directory, Gradle version, and JUnit extension context
+     * @param delegate the invoker to decorate
+     * @return the decorated invoker
+     */
```
CRogers: Fwiw, I think Javadoc like this doesn't really add much value over just the names of method and parameter. It does make it a bit harder to review as there's more stuff to read. The really valuable javadoc give you extra info you need to know, like the info on ordering at the class level.

CRogers: Like especially the `context` one here feels like it would get out of date super easily if we added something else to the context.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -27,18 +40,91 @@ public interface GradleInvoker {
...
+     *
+     * <p>Decorators are collected from the test extension context hierarchy, starting from the test method
+     * and moving up through parent classes. They are applied in registration order: first-registered decorators become
+     * innermost wrappers, while later-registered decorators become outer wrappers.
+     *
+     * @param path the root project directory
+     * @param gradleVersion the Gradle version to use
+     * @param extensionContext the JUnit extension context containing registered decorators
+     * @return a decorated GradleInvoker
+     */
```
CRogers: Again, not really sure how useful javadoc is on a non-public method - much of this information already exists on other classes. I would hope that people calling `GradleInvoker#create` wouldn't need to worry about decorators, the decorator behaviour should all be hidden inside.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
CRogers: I guess the `DecoratorFactory` could get a list of annotations rather than just one? Then it would handle repeated annotations, and each implementation could decide whether to throw an exception on repeated annotations or not, and how to combine multiple annotations. We do not need gradle-plugin-testing to handle or worry about conflicts then, it just finds all the relevant annotations.

The factory is kinda nice as it could handle these repeated annotations, but also allows you to configure the decorators with either parameters from the annotations or stuff from the `ExtensionContext`.

If we allowed multiple different annotation types to the same factory, perhaps this would allow us to get rid of the custom logic for adding the configuration cache annotation [here](https://github.com/palantir/gradle-plugin-testing/blob/ba579188245590c25dc47f7d16fecaeaab082aa7/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java#L64-L67). Instead of having a `ConfigurationCacheParameterExtension` and [a stateful store like we currently do](https://github.com/palantir/gradle-plugin-testing/blob/f0237d292dee4479eb27b0122681939705e4aa18/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheStore.java#L26), we could have something like this:

```java
@RegistersGradleInvokerDecorator(ConfigurationCacheDecoratorFactory.class)
@interface GradlePluginTests {}

@RegistersGradleInvokerDecorator(ConfigurationCacheDecoratorFactory.class)
@interface DisabledConfigurationCache {}

class ConfigurationCacheDecoratorFactory implements GradleInvokerDecoratorFactory {
    GradleInvokerDecorator create(GradleInvoker delegate, List<Object> annotations) {
         // ignore GradlePluginTests
         // see if DisabledConfigurationCache exists, if it does do not decorate (`return delegate`)
         // otherwise look for global junit parameter to decide whether we decorate or not 
    }
}
```

Imagine if we added `@WithConfigurationCache` (we opted not to do this, but imagine), we'd just need to add this and change the logic in the decorator factory:

```java
@RegistersGradleInvokerDecorator(ConfigurationCacheDecoratorFactory.class)
@interface WithConfigurationCache {}
```

note: This requires the declaration order is maintained in the annotation list (eg "outer" annotations before "inner" ones, eg test class  annotations first, then then test methods etc). Which is what I'd expect anyway.

What do you think @crogoz @felixdesouza?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -27,18 +48,106 @@ public interface GradleInvoker {
...
+    private static List<Annotation> findAllAnnotationsWithRegisterDecorator(AnnotatedElement element) {
+        ImmutableList.Builder<Annotation> annotationBuilder = ImmutableList.builder();
+
+        for (Annotation annotation : element.getAnnotations()) {
+            if (AnnotationSupport.isAnnotated(annotation.annotationType(), RegistersGradleInvokerDecorator.class)) {
+                annotationBuilder.add(annotation);
+            }
+            // Check if this might be a container annotation for repeatable annotations
+            // Look for annotations that might be containers for @RegistersGradleInvokerDecorator annotations
+            try {
```
crogoz, line 93: a `@Repeatable` annotation will group the repeated annotations in a container (eg. see the tests [here](https://github.com/palantir/gradle-plugin-testing/pull/344/changes#diff-d8c7f252099aef56377f42c812500fb471a780ea9003a4bfa4ad697f49a0f9e8R35)). 

I am using reflection to retrieve all the annotations inside the container that might be of type `RegistersGradleInvokerDecorator`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -27,18 +48,106 @@ public interface GradleInvoker {
...
+        // Reverse to get top-down order (parent classes -> class -> method)
+        Collections.reverse(contextHierarchy);
+
+        Stream<Annotation> classAnnotations = contextHierarchy.stream()
+                .flatMap(ctx -> ctx.getTestClass().stream())
+                .flatMap(testClass -> findAllAnnotationsWithRegisterDecorator(testClass).stream());
+
+        return Stream.concat(classAnnotations, methodAnnotations.stream())
+                // preserving the order while dropping duplicated annotations
+                .collect(Collectors.toCollection(LinkedHashSet::new));
```
crogoz, line 81: there are some duplicated annotations (eg. ClassTemplateExtensionContext and TestClassExtensionContext), so I need to make sure they are only applied once. 

added logic to add the decorators from top to bottom (outer classes -> test methods)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorDiscoveryExtension.java
crogoz: let me know if this is what you had in mind https://github.com/palantir/gradle-plugin-testing/pull/344/changes/8a8c31e860033a1bc4c30aa039203610ab34c4d6

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecoratorFactory.java
```diff
@@ -0,0 +1,61 @@
...
+import com.palantir.gradle.testing.execution.GradleInvoker;
+import java.lang.annotation.Annotation;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Factory for creating {@link ConfigurationCacheDecorator} instances.
+ */
+public final class ConfigurationCacheDecoratorFactory implements GradleInvokerDecoratorFactory {
```
CRogers: Really nice all the configuration cache logic is just in this class now and no extra extensions/store/GradleInvoker modifications are needed.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorFactory.java
```diff
@@ -0,0 +1,41 @@
...
+ * Factory for creating {@link GradleInvokerDecorator} instances from annotations.
+ *
+ * <p>Implementations are referenced by the {@link RegistersGradleInvokerDecorator} meta-annotation.
+ * When a test class or method is annotated with an annotation that has {@code @RegistersGradleInvokerDecorator},
+ * the framework will instantiate the factory and call {@link #create} with all matching annotation instances.
+ *
+ * <p>Factories must have a public no-argument constructor.
+ *
+ */
+public interface GradleInvokerDecoratorFactory {
```
CRogers: I thought we'd have to get rid of the generic here if we wanted to support multiple annotation types. This means it's slightly more annoying to use if you only have one annotation type. But I don't think this is the case; we could have a bound `GradleInvokerDecoratorFactory<A extends Annotation>` so if there's one annotation type you can specify it directly:

```java
class JdkAutomanagementDecoratorFactory extends GradleInvokerDecoratorFactory<WithJdkAutomanagement> {
```

vs multiple annotation types use `Annotation` as the bound:

```java
class ConfigurationCacheDecoratorFactory extends GradleInvokerDecoratorFactory<Annotation> {
```

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
```diff
@@ -0,0 +1,104 @@
...
+import com.palantir.gradle.testing.execution.GradleInvoker;
+import com.palantir.gradle.testing.execution.InvocationResult;
+import com.palantir.gradle.testing.junit.TestDecorators.WithArgAddingDecorator;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@WithArgAddingDecorator(arg = "-PtopClass=true")
+@WithArgAddingDecorator(arg = "-PclassValue=classDecorator")
```
CRogers: Nice to not need multiple of this same annotation now!

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
```diff
@@ -0,0 +1,104 @@
...
+@GradlePluginTests
+@DisabledConfigurationCache("Testing decorator mechanism without config cache complexity")
+class GradleInvokerDecoratorTests {
+
+    @BeforeEach
+    void setUpWithAGradleInvokerWorks(GradleInvoker _invoker, RootProject rootProject) {
```
CRogers, line 62: I think we should test `GradleInvoker` decoration works in a `@BeforeEach`. Probably not this one - can make a nested class with one test (with no method body) to add a standalone `@BeforeEach` without affecting these other tests.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
```diff
@@ -0,0 +1,104 @@
...
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@WithArgAddingDecorator(arg = "-PtopClass=true")
+@WithArgAddingDecorator(arg = "-PclassValue=classDecorator")
+@GradlePluginTests
+@DisabledConfigurationCache("Testing decorator mechanism without config cache complexity")
+class GradleInvokerDecoratorTests {
```
CRogers, line 35: I think in this test we only check that `@Repeatable` annotations work. Probably good to test a non `@Repeatable` annotation too.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
CRogers: ~~Also good to test `@Nested` classes~~ Ah, I missed the existing `@Nested` test

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
felixdesouza: do you still want to do this here? or put it in a decorator :P

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
CRogers: I wonder if we can use `AnnotationSupport. findRepeatableAnnotations` if we mark `@RegistersGradleInvokerDecorator` as `@Inheritable`?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -27,18 +48,107 @@ public interface GradleInvoker {
...
+    }
+
+    private static List<Annotation> findAllAnnotationsWithRegisterDecorator(AnnotatedElement element) {
+        ImmutableList.Builder<Annotation> annotationBuilder = ImmutableList.builder();
+
+        for (Annotation annotation : element.getAnnotations()) {
```
felixdesouza, line 87: I feel like you don't have to do a lot of this?

`AnnotationSupport.isAnnotated` finds annotations if they're directly present, in a superclass/interface, and then also if they're metapresent i.e. present on annotations on the . 

I'm not sure how that then plays into repeatable annotations yet, haven't finished reading the code, but it feels like you shouldn't have to do a lot of this perhaps? Easier to validate this with a unit test I suppose.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecoratorFactory.java
felixdesouza: yes, this looks really good!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecoratorFactory.java
```diff
@@ -0,0 +1,61 @@
...
+    public GradleInvokerDecorator create(List<Annotation> annotations) {
+        boolean disabledConfigurationCachePresent =
+                annotations.stream().anyMatch(annotation -> annotation instanceof DisabledConfigurationCache);
+
+        if (disabledConfigurationCachePresent) {
+            log.debug("DisabledConfigurationCache annotation found, skipping configuration cache decoration");
+            return (context, delegate) -> delegate;
+        }
+
+        return (context, delegate) -> {
```
felixdesouza: chatting with @CRogers, it's not entirely clear that `GradleInvokerDecorator` and `GradleInvokerDecoratorFactory` need to be separate types. The type specific decorator ends up being super simple. You could literally just tack on an `.apply()` and then return. 

I'm wondering whether we collapse the type, and if we *do* need to have a factory at a later date, it should be relatively easy to do.

Then adding a decorator is just:
* Adding an invoker
* Adding a decorator

If you wanted to have a factory at a later date, you could get the decorators registered in this "new way" and then have a trivial factory, and still kinda works just fine.

I'm mostly struggling to distinguish between what code lives in the decorator vs what lives in the factory for example.

Potentially worth playing around with this and seeing what falls out?

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/TestDecorators.java
```diff
@@ -0,0 +1,68 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Stream;
+
+public class TestDecorators {
```
felixdesouza, line 27: yeah, this is a really good canonical example of what's required to add a new decorator, and it's quite succinct!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
CRogers: The difficultly is we need the annotations that the `@RegistersGradleInvokerDecorator` is on, not just all the `@RegistersGradleInvokerDecorator`. This means you can't directly use the junit `AnnotationSupport` methods. Dealing with repeatable annotations is also incredibly annoying, but I don't think it's possible to do better than reading the `value` method using reflection as @crogoz has done - I couldn't find any small library that would handle this easily either - just spring-core, which is overkill.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
CRogers: No, see https://github.com/palantir/gradle-plugin-testing/pull/344#discussion_r2742883449

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorFactory.java
crogoz: I have to use reflection for this - if we only want to pass in the annotation types the factory supports - https://github.com/palantir/gradle-plugin-testing/pull/344/commits/9bc3c753eeea7a72c73f207e209dfdc5e023d940. I am not sure if there is another way to do this.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecoratorFactory.java
crogoz: changed the logic in https://github.com/palantir/gradle-plugin-testing/pull/344/commits/0f058db9b9cf909f21fc5741e4fe1ab7e28c33f8 for dropping the factory, and I only kept the decorator & the invoker, let me know what you think

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecorator.java
```diff
@@ -0,0 +1,58 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class ConfigurationCacheDecorator implements GradleInvokerDecorator<DisabledConfigurationCache> {
+
+    private static final Logger log = LoggerFactory.getLogger(ConfigurationCacheDecorator.class);
+
+    @Override
+    public GradleInvoker decorate(
+            DecoratorContext context, GradleInvoker delegate, List<DisabledConfigurationCache> annotations) {
```
CRogers: How does this work? We register the decorator on `GradlePluginTests` too? I would have thought we'd have to make this a `Annotation` bound.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,29 +16,166 @@
...
+                if (GradleInvokerDecorator.class.equals(parameterizedType.getRawType())) {
+                    Type typeArg = parameterizedType.getActualTypeArguments()[0];
+                    if (typeArg instanceof Class<?> annotationClass) {
+                        return (Class<? extends Annotation>) annotationClass;
+                    }
+                }
+            }
+        }
+
+        // If we couldn't find it directly, check the superclass
```
CRogers: If we'd really want to be correct, we'd explore up every superclass and superinterface, not just the superclass.... I think it's relatively succinct to write this using streams, I've done it a couple of times, might be a library.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecorator.java
crogoz: > How does this work? We register the decorator on GradlePluginTests too?

I am registering the decorator in GradlePluginTests [here](https://github.com/palantir/gradle-plugin-testing/pull/344/changes#diff-491ee1c9d357eb85b05857bf940e5777f0390b77c3388b2e89bb0d928bc6f4deR32). 

> I would have thought we'd have to make this a Annotation bound.

I am passing now to the decorator when I instantiate it all the annotations that are of the templated type (eg. in this case `DisabledConfigurationCache` which means that this decorator knows how to work with/cares only about `DisabledConfigurationCache` annotations. [here](https://github.com/palantir/gradle-plugin-testing/pull/344/changes/9bc3c753eeea7a72c73f207e209dfdc5e023d940#diff-63221b4e27560e48da5d0ad3e2ff056196ba2914277ea6c9a134213432a2085cR141-R147) I am populating the annotations only with the ones we support.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
```diff
@@ -0,0 +1,142 @@
...
+
+            InvocationResult result = invoker.withArgs("hello").buildsSuccessfully();
+            result.assertThat()
+                    .output()
+                    .contains("Hello from task hello with decorator values: classDecorator & secondClassDecorator")
+                    .contains("foo bar baz");
+            result.assertThat().output().contains("Task :help");
+
+            // top-down order (parent classes -> class -> method)
+            assertThatThrownBy(() -> invoker.withArgs("hello").buildsWithFailure())
```
felixdesouza: nit, put the comment as an `.as` pls

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/GradleInvokerDecoratorTests.java
```diff
@@ -0,0 +1,142 @@
...
+                    }
+                }
+                """);
+
+            InvocationResult result = invoker.withArgs("hello").buildsSuccessfully();
+            result.assertThat()
+                    .output()
+                    .contains("Hello from task hello with decorator values: classDecorator & secondClassDecorator")
+                    .contains("foo bar baz");
+            result.assertThat().output().contains("Task :help");
```
felixdesouza: combine this with the above?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,29 +16,166 @@
...
+
+        for (Annotation annotation : element.getAnnotations()) {
+            if (AnnotationSupport.isAnnotated(annotation.annotationType(), RegistersGradleInvokerDecorator.class)) {
+                annotationBuilder.add(annotation);
+            }
+            // Check if this might be a container annotation for repeatable annotations
+            // Look for annotations that might be containers for @RegistersGradleInvokerDecorator annotations
+            try {
+                Method valueMethod = annotation.annotationType().getDeclaredMethod("value");
+                if (valueMethod.getReturnType().isArray()) {
```
felixdesouza: nit: do you mind inverting the ifs here? it'll read a bit more linearly and less indentation 🙏

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecorator.java
CRogers: This feels confusing to me - I think I'd rather just give every single annotation with `@RegistersGradleInvokerDecorator(@ThisDecorator)` rather than filtering. I know this whole thing is magic, but this filtering feels especially magic.

CRogers: I think I'd rather it just threw an exception if the annotation wasn't in the right bound.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,29 +16,166 @@
...
+
+    /**
+     * Determines the annotation type that a decorator can process by examining its generic type parameter.
+     */
+    @SuppressWarnings("unchecked")
+    private static Class<? extends Annotation> getDecoratorAnnotationType(
+            Class<? extends GradleInvokerDecorator> decoratorClass) {
+        // Look for the GradleInvokerDecorator interface in the class hierarchy
+        Type[] genericInterfaces = decoratorClass.getGenericInterfaces();
+        for (Type genericInterface : genericInterfaces) {
```
felixdesouza: similarly here:

```java
for (Type genericInterface : genericInterfaces) {
    if (!(genericInterface instanceof ParameterizedType parameterizedType)) {
        continue;
    }
    
    if (!GradleInvokerDecorator.class.equals(parameterizedType.getRawType())) {
        continue;
    }
    
    Type typeArg = parameterizedType.getActualTypeArguments()[0];
    if (typeArg instanceof Class<?> annotationClass) {
        return (Class<? extends Annotation>) annotationClass;
    }
}
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecorator.java
crogoz: changed the logic [here](https://github.com/palantir/gradle-plugin-testing/pull/344/commits/318a85e5e9324c94e11cd6a2e46928c1569068c5)  and added logic to throw an exception when the types don't match [here](https://github.com/palantir/gradle-plugin-testing/pull/344/commits/c34b502812067993c196e0d010e6cd2ab9d51b12).

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,29 +16,185 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    private static GradleInvoker createGradleInvokerFromDecorator(
+            Class<? extends GradleInvokerDecorator> decoratorClass,
+            DecoratorContext context,
+            GradleInvoker invoker,
+            List<Annotation> annotations) {
+        try {
+            checkAnnotationsType(decoratorClass, annotations);
```
crogoz, line 151: without this, we would get either an error of this form: 

```
class jdk.proxy3.$Proxy24 cannot be cast to class com.palantir.gradle.testing.junit.TestDecorators$WithArgAddingDecorator (jdk.proxy3.$Proxy24 is in module jdk.proxy3 of loader 'app'; com.palantir.gradle.testing.junit.TestDecorators$WithArgAddingDecorator is in unnamed module of loader 'app')
java.lang.ClassCastException: class jdk.proxy3.$Proxy24 cannot be cast to class com.palantir.gradle.testing.junit.TestDecorators$WithArgAddingDecorator (jdk.proxy3.$Proxy24 is in module jdk.proxy3 of loader 'app'; com.palantir.gradle.testing.junit.TestDecorators$WithArgAddingDecorator is in unnamed module of loader 'app')
``` 
(this is when I tried to add the wrong decorator to one of the tests in the `GradleInvokerDecoratorTests`)

or 

```
org.junit.jupiter.api.extension.ParameterResolutionException: Failed to resolve parameter [com.palantir.gradle.testing.execution.GradleInvoker invoker] in method [void com.palantir.example.IncompatibleDecoratorsFixtureTest.some_test(com.palantir.gradle.testing.execution.GradleInvoker,com.palantir.gradle.testing.project.RootProject)]: Failed to instantiate decorator class SomeDecorator
...
Caused by: java.lang.IllegalAccessException: interface com.palantir.gradle.testing.execution.GradleInvoker cannot access a member of class com.palantir.example.SomeDecorator with modifiers
```
(using the FixtureTest)

None of them don't give too much info about the actual reason (annotation type mismatch).

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,29 +16,186 @@
...
+        try {
+            Method decorateMethod =
+                    decoratorClass.getMethod("decorate", DecoratorContext.class, GradleInvoker.class, List.class);
+            Type[] genericParameterTypes = decorateMethod.getGenericParameterTypes();
+            if (genericParameterTypes.length >= 3 && genericParameterTypes[2] instanceof ParameterizedType paramType) {
+                Type[] typeArgs = paramType.getActualTypeArguments();
+                if (typeArgs.length > 0 && typeArgs[0] instanceof Class) {
+                    return (Class<?>) typeArgs[0];
+                }
             }
```
felixdesouza, line 192: so if I've understood correctly, this is getting the type `T` of `List<T>` from the third parameter?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecorator.java
```diff
@@ -0,0 +1,59 @@
...
+
+public final class ConfigurationCacheDecorator implements GradleInvokerDecorator<Annotation> {
+
+    private static final Logger log = LoggerFactory.getLogger(ConfigurationCacheDecorator.class);
+
+    @Override
+    public GradleInvoker decorate(DecoratorContext context, GradleInvoker delegate, List<Annotation> annotations) {
+        if (annotations.stream()
+                .anyMatch(
+                        annotation -> annotation.annotationType().isAssignableFrom(DisabledConfigurationCache.class))) {
```
CRogers: nit: Annotations can't extend each other so it's probably good enough to do `.equals`.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
crogoz: yes, I'll add some comments about this

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/RegistersGradleInvokerDecorator.java
```diff
@@ -0,0 +1,54 @@
...
+import java.lang.annotation.Target;
+import java.util.List;
+
+/**
+ * Meta-annotation that marks an annotation as registering a {@link GradleInvokerDecorator}.
+ *
+ * <p>When a test class or method is annotated with an annotation that has this meta-annotation,
+ * the framework will automatically:
+ * <ol>
+ *   <li>Instantiate the specified {@link GradleInvokerDecorator}</li>
```
felixdesouza, line 32: we should be clearer about the contract i.e. it's a no arg constructor

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,29 +16,186 @@
...
+
+                if (!Annotation.class.isAssignableFrom(componentType)) {
+                    continue;
+                }
+                if (!componentType.isAnnotationPresent(RegistersGradleInvokerDecorator.class)) {
+                    continue;
+                }
+                Object value = valueMethod.invoke(annotation);
+                if (value instanceof Annotation[] containedAnnotations) {
+                    annotationBuilder.addAll(Arrays.stream(containedAnnotations).toList());
```
felixdesouza: nit, I think you can just do `annotationBuilder.add(containedAnnotations)`  instead of `addAll` without having to convert it to a `List` first.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -16,36 +16,192 @@
...
+    }
+
+    /**
+     * Groups annotations by their decorator class and decorates the baseInvoker passing the relevant
+     * `@RegistersGradleInvokerDecorator` annotations.
+     */
+    private static GradleInvoker decorateInvokerWithAnnotations(
+            DecoratorContext context, GradleInvoker baseInvoker, Set<Annotation> annotations) {
+
+        Map<Class<? extends GradleInvokerDecorator>, List<Annotation>> decoratorToAnnotations = new LinkedHashMap<>();
```
CRogers: It would be best to use a `ListMultimap` here if we access to guava. No need for the if check below then.

CRogers (approved):

crogoz:
👍

PR #353, opened 2026-01-23
Add migration gotchas and clarify external plugin usage

felixdesouza:
## Before this PR
The testing guide's "Testing with External Plugins" section doesn't clarify that `gradlePluginForTesting` is already created by the plugin, and the example uses an explicit version number.

Migration-specific gotchas were scattered or missing from documentation.

## After this PR

**testing-guide.md (general usage):**
- Clarify that `gradlePluginForTesting` configuration is automatically created by the plugin
- Update example to show multiple dependencies without explicit versions
- Add guidance on chaining `.add()` calls for plugins

**junit-migration-excavator-instructions.md (migration-specific):**
- Add guidance: do not add `gradle-plugin-testing-junit` to `versions.props` as the plugin provides it automatically
- Add "Don't Use buildscript Blocks for External Plugins" section
- Add "Don't Access Framework Internals" section (no need for version-based namespacing)
- Add "Use Full Task Paths for Explicit Assertions" section

Based on issues discovered during review of migration PR https://github.com/palantir/gradle-failure-reports/pull/432

## Possible downsides?
None - documentation only.

### DIFF
--- a/docs/testing-guide.md
+++ b/docs/testing-guide.md
@@ -37,6 +37,7 @@ This guide covers how to write tests for Gradle plugins using the `gradle-plugin
 - [Error Prone Checks](#error-prone-checks)
     - [Why We Ship Error Prone Checks](#why-we-ship-error-prone-checks)
     - [Configuring Error Prone](#configuring-error-prone)
+    - [Common Migration Gotchas](#common-migration-gotchas)
 
 ## Quick Start
 
@@ -256,7 +257,7 @@ Use the structured `plugins()` API to add plugins:
 ```java
 @Test
 void add_plugins(RootProject project) {
-    // Add plugins individually
+    // Chain multiple plugins together for readability
     project.buildGradle()
         .plugins()
         .add("java")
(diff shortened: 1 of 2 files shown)


docs/testing-guide.md
```diff
@@ -751,3 +782,37 @@ plugins {
 ```
 
 The framework's Error Prone checks will be automatically registered and applied to your test code, catching issues at compile time and offering automated fixes where possible.
+
+### Common Migration Gotchas
+
+When migrating from Nebula/Spock tests to this framework, avoid these common mistakes:
```
felixdesouza: should this live in the `junit-migration-excavator-instructions` file instead?

FinlayRJW: maybe? feels more AI focused so maybe should be moved?

docs/testing-guide.md
```diff
@@ -269,17 +270,29 @@ void add_plugins(RootProject project) {
 }
 ```
 
+When adding multiple plugins, chain the `.add()` calls rather than calling `.plugins()` multiple times:
+
+```java
+// Preferred - chain the calls
+rootProject.buildGradle().plugins().add("com.palantir.failure-reports").add("java");
```
CRogers: Is this better? I think it's fine if the `.add`s are on multiple lines, but I think the `.add`s on the same line is less readable.

docs/testing-guide.md
CRogers: Yeah, I don't think this file should reference migrations. Hopefully people can just read it without having to worry about groovy based tests at some point.

docs/testing-guide.md
```diff
@@ -751,3 +782,37 @@ plugins {
...
+
+When migrating from Nebula/Spock tests to this framework, avoid these common mistakes:
+
+#### Don't Access Framework Internals
+
+The framework isolates each test run into its own directory, namespaced by Gradle version. You should never need to know which Gradle version is running.
+
+```java
+// WRONG - Accessing implementation details
+String version = ((DefaultGradleInvoker) gradle).gradleVersion().version();
```
CRogers: Maybe we should be more liberal with `@RestrictedApi` too if these need to be public. I kinda hate it but we probably just have put everything in same package.

felixdesouza: for restricted api absolutely, there's no downside and it guides the ai more I believe?

docs/testing-guide.md
felixdesouza: that's fair, but it's also up to formatting. I wonder whether a varargs thing would be better

CRogers (approved):

PR #358, opened 2026-01-28
Add `@ParameterizedByGradleVersion` for version-conditional test parameters

FinlayRJW:
## Before this PR
When testing Gradle plugins across multiple versions, tests often need different behavior based on the Gradle version. Previously, this required maintaining separate test methods for different version ranges.

## After this PR

Introduces `@ParameterizedByGradleVersion` - an annotation that injects a String parameter based on the Gradle version under test using a condition-based approach with a default fallback.

**Usage (single parameter):**
```java
@ParameterizedByGradleVersion(
    when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
    otherwiseString = "new")
void test(GradleInvoker gradle, RootProject project, @InjectByGradleVersion String behavior) {
   // For Gradle < 8.0: behavior = "old"
   // For Gradle >= 8.0: behavior = "new"
}
```

**Multiple conditions** for finer control:
```java
@ParameterizedByGradleVersion(
    when = {
        @WhenVersion(lessThan = "8.0", stringValue = "less than 8"),
        @WhenVersion(lessThan = "9.0", stringValue = "8.x")
    },
    otherwiseString = "9 and up")
void test(GradleInvoker gradle, RootProject project, @InjectByGradleVersion String behavior) {
   // behavior = "less than 8" for Gradle < 8.0
   // behavior = "8.x" for Gradle 8.0 to 8.x
   // behavior = "9 and up" for Gradle >= 9.0
}
```

**Multiple parameters** with different version conditions (name required):
```java
@ParameterizedByGradleVersion(
    name = "style",
    when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
    otherwiseString = "new")
@ParameterizedByGradleVersion(
    name = "format",
    when = @WhenVersion(lessThan = "9.0", stringValue = "classic"),
    otherwiseString = "modern")
void test(GradleInvoker gradle, @InjectByGradleVersion String style, @InjectByGradleVersion String format) {
   // Each parameter resolved independently based on its conditions
}
```

**Works with `@BeforeEach`:**
```java
@BeforeEach
@ParameterizedByGradleVersion(
    when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
    otherwiseString = "new")
void setup(RootProject project, @InjectByGradleVersion String behavior) {
    // behavior resolved based on the Gradle version under test
}
```

### Validation

- `@WhenVersion` conditions must be ordered by ascending `lessThan` version (lowest first)
- Duplicate `name` values on the same method are not allowed
- Same `name` can be used across different methods (e.g., `@BeforeEach` and `@Test`)
- When multiple `@ParameterizedByGradleVersion` annotations are present, all must have `name` specified

### Key Design Decisions

**Condition-based with `otherwiseString` fallback:** Uses `@WhenVersion(lessThan = "X")` conditions checked in order, with `otherwiseString` as the default. This is simpler than requiring contiguous ranges that cover all versions.

**`@InjectByGradleVersion` marks receiving parameters:** The receiving String parameter must be annotated with `@InjectByGradleVersion`. For single annotations, the `name` is optional. For multiple annotations, `name` is required and must match the parameter name.

**Works with `@BeforeEach` and `@Test`:** The resolver uses `parameterContext.getDeclaringExecutable()` to get annotations from the actual method being invoked, so it works with lifecycle methods.

**Repeatable for multiple parameters:** Multiple annotations with different `name` values can inject different version-conditional values into different parameters.

## Possible downsides?

- Only supports String parameters - other types would need additional resolver logic


docs/testing-guide.md
```diff
@@ -171,6 +172,38 @@ class CompatibilityTest {
...
+    project.buildGradle().append("myPlugin.style = '%s'", configOption);
+}
+```
+
+**Version conditions** (use exactly one per `@WhenVersion`):
+- `lessThan = "8.0"` - versions below 8.0
+- `lessThanOrEqualTo = "8.0"` - versions up to and including 8.0
+- `equalTo = "8.0"` - exactly version 8.0
+
+**Multiple annotations** create a Cartesian product. Values inject positionally into String parameters:
```
CRogers: Making it the cartesian product seems very weird to me? We have a fixed number of Gradle versions we're testing against, shouldn't we just use the right value of each parameter for each Gradle version? The combinatorial explosion of cartesian product doesn't seem to do what we want?

docs/testing-guide.md
```diff
@@ -171,6 +172,38 @@ class CompatibilityTest {
...
+
+```java
+@ParameterizedByGradleVersion(value = @WhenVersion(lessThan = "8.0", value = "legacy"), otherwise = "new")
+@ParameterizedByGradleVersion(value = @WhenVersion(lessThan = "8.0", value = {"1", "2"}), otherwise = "4")
+void test(GradleInvoker gradle, RootProject project, String feature, String maxWorkers) {
```
CRogers: How do we know that the `feature` parameter corresponds to the first `@ParameterizedByGradleVersion`? Is it just order? I'd think it'd make sense to have an annotation on the parameter linking it somehow.

CRogers: Perhaps we could even put the whole `@ParameterizedByGradleVersion` on the parameter? Especially if we use a `ParameterResolver` for this?

FinlayRJW: currently just order, top is first think this is how valueSource works - can add a name param to `ParameterizedByGradleVersion` to set the var name if this is better?

docs/testing-guide.md
```diff
@@ -171,6 +172,38 @@ class CompatibilityTest {
...
+    // Gradle < 8.0: (legacy, 1), (legacy, 2)
+    // Gradle >= 8.0: (new, 4)
+}
+```
+
+> **Note:** Do not combine with `@Test` - the annotation includes `@TestTemplate`.
```
CRogers: Is this a test template because of the cartesian product behaviour? I think it's a bit strange - I think we could just write this as a `ParameterResolver` without it?

docs/testing-guide.md
FinlayRJW: > Perhaps we could even put the whole @ParameterizedByGradleVersion on the parameter

Hmm this could end up with some odd formatting? 

```java
    void test_one(GradleInvoker gradleInvoker, RootProject rootProject, String behavior,
                  @ParameterizedByGradleVersion(
            otherwise = {"3", "4"},
            value =
            @WhenVersion(
                    lessThan = "9.3.0",
                    value = {"1", "2"})) maxInt) {
```

docs/testing-guide.md
```diff
@@ -171,6 +172,38 @@ class CompatibilityTest {
...
+**Version conditions** (use exactly one per `@WhenVersion`):
+- `lessThan = "8.0"` - versions below 8.0
+- `lessThanOrEqualTo = "8.0"` - versions up to and including 8.0
+- `equalTo = "8.0"` - exactly version 8.0
+
+**Multiple annotations** create a Cartesian product. Values inject positionally into String parameters:
+
+```java
+@ParameterizedByGradleVersion(value = @WhenVersion(lessThan = "8.0", value = "legacy"), otherwise = "new")
+@ParameterizedByGradleVersion(value = @WhenVersion(lessThan = "8.0", value = {"1", "2"}), otherwise = "4")
```
CRogers: It seems super strange that we're allowing people to have multiple values for each Gradle version. The only reason I could think to do this is to combine functionality from `@ParameterisedTest`, but I'm not sure that's ever going to be needed. Can you give an example?

docs/testing-guide.md
CRogers: I suspect using `@TestTemplate` will prevent people from using `@ParameterizedTest` - part of the reason we waited until `@ClassTemplate` existed rather than using `@TestTemplate` to test against multiple gradle versions.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(ParameterizedByGradleVersions.class)
+@TestTemplate
+@ExtendWith(ParameterizedByGradleVersionTestTemplateProvider.class)
+public @interface ParameterizedByGradleVersion {
+
+    /** Version-specific value conditions. */
+    WhenVersion[] value();
```
CRogers: I would give this a more descriptive name like `when`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /** Default value(s) when no version condition matches. */
+    String[] otherwise() default {};
```
CRogers: Won't we always need an `otherwise`? The `WhenVersion`s are always going to `<`/`<=`/`==` bounds, so when a new version of Gradle is released that is higher than these bounds, suddenly we'll have no value for that parameter?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WhenVersion.java
```diff
@@ -0,0 +1,43 @@
...
+    String equalTo() default "";
+
+    /** Matches versions below this (e.g., "9.0" matches 8.x). */
+    String lessThan() default "";
+
+    /** Matches versions up to and including this. */
+    String lessThanOrEqualTo() default "";
+
+    /** Values to use when this condition matches. */
+    String[] value() default {};
```
CRogers: Call this `stringValue` so we can do `int` etc later without it being named inconsistently

docs/testing-guide.md
CRogers: Honestly the formatting on the method level is already pretty strange in the tests - we may be able to fix the formatter.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
CRogers: The only reason to ever call an annotation param `value` is so you can can just elide the parameter name eg `@ParameterizedByGradleVersion(@WhenVersion(...`. But as I mentioned below I think we'll always need `otherwise`, so no real value?

PR.md
CRogers, line 1: Did you mean to commit this?

FinlayRJW: nope - removed

docs/testing-guide.md
FinlayRJW: I wrote this as a `TestTemplate` so that we could parameterise per test, when I tried to make this as a `ParameterResolver` I ended up having to create many tests then use a condition to skip tests that where not supposed to run - which made the output very noisy

docs/testing-guide.md
FinlayRJW: Hmm I was mostly thinking of if you had two params that should change with gradle version i.e. two paths that are different from 8 -> 9 - not sure if this would ever happen though

docs/testing-guide.md
felixdesouza: strong agree, I would actually even go further and ensure that if we're using this annotation, you cover the entire state space i.e. it is an error to have a gap in your version ranges.

I don't think we want to have this as a general skipping mechanism either, merely just the ability to provide a parameter based on a particular version range of gradle.

FinlayRJW: I did this for the case of `@ParameterizedByGradleVersion(value = @WhenVersion(lessThan = "8.0", value = {"1", "2"}), otherwise = "4")` i.e. a single `@WhenVersion` providing a set of tests `value = {"1", "2"}` if we are happy to restrict each annotation to provide a single value per version I think we can remove it

felixdesouza:
do we actually need `@WhenVersion` ? Ergonomically, this does not feel that nice, especially for the simple cases.

My guess is that it's complicated by the use of the cartesian product which I don't think we should do. If you flatten it, i.e. `@ParameterizedByGradleVersion` contains the when version inside, then repeating it is sufficient.

e.g. 

```java
@ParameterizedByGradleVersion(lessThan = "9.0", stringValue = "for-gradle-9")
@ParameterizedByGradleVersion(lessThan = "8.0", stringValue = "for-gradle-8")
@Test
void test(String parameter) {
 ...
}
``` 

If you do need the cross product at some point, I think there are probably other ways we could think about it, without making the simple cases ugly.

docs/testing-guide.md
FinlayRJW: > Hmm I was mostly thinking of if you had two params that should change with gradle version i.e. two paths that are different from 8 -> 9 - not sure if this would ever happen though

I suppose this could still work if we say one value per paramater per gradle version

FinlayRJW:
> do we actually need `@WhenVersion` ? Ergonomically, this does not feel that nice, especially for the simple cases.
> 
> My guess is that it's complicated by the use of the cartesian product which I don't think we should do. If you flatten it, i.e. `@ParameterizedByGradleVersion` contains the when version inside, then repeating it is sufficient.
> 
> e.g.
> 
> ```java
> @ParameterizedByGradleVersion(lessThan = "9.0", stringValue = "for-gradle-9")
> @ParameterizedByGradleVersion(lessThan = "8.0", stringValue = "for-gradle-8")
> @Test
> void test(String parameter) {
>  ...
> }
> ```
> 
> If you do need the cross product at some point, I think there are probably other ways we could think about it, without making the simple cases ugly.

Hmm this would restrict you to having one variable that can change with gradle version right? Unless we add a fieldName or something?

docs/testing-guide.md
felixdesouza: I would have thought we'd need to inject it/do the wiring at the stage where we create the templates for each `GradleVersion` ? It might not be immediately injectable, but then if you store it in context, then a parameter resolver could fetch it and inject it then. Then you have 1-1 mappings, and you don't break `@Test` or parameterized test.

docs/testing-guide.md
felixdesouza: I think we should keep the parameter definitions/declarations up top, and have it repeatable, see my below comment. Having the definitions on the parameter is pretty unsightly.

Then have an annotation that tells the `ParameterResolver` what to inject. e.g. 

```java
@ParameterizedByGradleVersion(...)
@ParameterizedByGradleVersion(...)
void test_one(GradleInvoker gradleInvoker, RootProject rootProject, @ParameterInject String behaviour) {

}
```

I don't really care for the name of this annotation, but as long as it's something that makes sense, I'm easy.

If we do want to have multiple parameters/cross product behaviour or what not - which I think we should do when the need arises - we can use some sort of named property. Then determine the semantics at a later point.

```java
@ParameterizedByGradleVersionGroup(
  name = "parameter1",
  versions = {
      @ParameterizedByGradleVersion(...),
      @ParameterizedByGradleVersion(...)
  }
)
@ParameterizedByGradleVersionGroup(
  name = "parameter2",
  versions = {
      @ParameterizedByGradleVersion(...),
      @ParameterizedByGradleVersion(...)
  }
)
void test_one(GradleInvoker gradleInvoker, RootProject rootProject, @Inject @Named("parameter1) String behaviour, @ParameterInject @Named("parameter2") String anotherThing) {
  // ...
}
```

We could probably get rid of the `@Named` annotation, and just ensure that the name of the group, corresponds to the parameter name or something.

I'm not suggesting we do this now as the PR is already large, and I'm not even sure if it's required. But my feeling is that it would be better if the annotations were composable, rather than we have to specify a bunch of things for the simple case => it is easier/nicer to just keep the test the same.

felixdesouza:
> > do we actually need `@WhenVersion` ? Ergonomically, this does not feel that nice, especially for the simple cases.
> > My guess is that it's complicated by the use of the cartesian product which I don't think we should do. If you flatten it, i.e. `@ParameterizedByGradleVersion` contains the when version inside, then repeating it is sufficient.
> > e.g.
> > ```java
> > @ParameterizedByGradleVersion(lessThan = "9.0", stringValue = "for-gradle-9")
> > @ParameterizedByGradleVersion(lessThan = "8.0", stringValue = "for-gradle-8")
> > @Test
> > void test(String parameter) {
> >  ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If you do need the cross product at some point, I think there are probably other ways we could think about it, without making the simple cases ugly.
> 
> Hmm this would restrict you to having one variable that can change with gradle version right? Unless we add a fieldName or something?

I think this is a fine restriction, and we can open it up to something else if the need arises, but I think making it simple to use for simple cases is more valuable imo. See my above [comment](https://github.com/palantir/gradle-plugin-testing/pull/358#discussion_r2736491173) on how we can evolve this if the need arises.

docs/testing-guide.md
FinlayRJW: Okay so after having a chat in-person I think it seems sensible that for now we support a single injected param per gradle version, the user has to upfront cover the entire potential version space so the annotation would look something like:

```java
@ParameterizedByGradleVersion(upperBound = "8.0", stringValue = "less than 8")
@ParameterizedByGradleVersion(lowerBound = "8.0", upperBound = "9.0", stringValue = "all 8.x")
@ParameterizedByGradleVersion(lowerBound = "9.0", stringValue = "9 and up")
void test_one(GradleInvoker gradleInvoker, RootProject rootProject, @ParameterInject String behaviour) {

}
``` 

where `lowerBound` is inclusive and `upperBound` is exclusive. I think this makes more sense than overlapping `lessThans` as its not immediately clear what parameter will run for each gradle version

Does this sound sensible to you @CRogers ?

docs/testing-guide.md
```diff
@@ -171,6 +172,38 @@ class CompatibilityTest {
...
+}
+```
+
+**With multiple ranges:**
+
+```java
+@Test
+@ParameterizedByGradleVersion(upperBound = "8.0", stringValue = "legacy")
+@ParameterizedByGradleVersion(lowerBound = "8.0", upperBound = "9.0", stringValue = "8.x")
+@ParameterizedByGradleVersion(lowerBound = "9.0", stringValue = "modern")
```
CRogers: Bit unsure about this new approach, for two reasons:
1. The `upper`/`lower` bound part feels duplicative and I think the previous approach was a bit easier to follow as it looks a bit like an if else tree:
   ```java
    if (gradleVersion < "8.0") { ... }
    else if (gradleVersion < "9.0") { ... }
   ```
2. Now we can only have one parameter per method. This feels strange to me - what happens if you want more than one? I kinda liked that you could have multiple parameters, and each were contained to their own outer annotation.

docs/testing-guide.md
CRogers: We really need github-notifications

docs/testing-guide.md
CRogers: Oh, I see there was a conversation with @felixdesouza outside the diff (fwiw, I don't like PR level comments as there's no way to reply directly to them).

FinlayRJW: I worry the if else tree approach is a bit ambiguous - do we always go from the first annotation down? or do we match to the closest version constraint i.e. if a user writes:

 ```
@ParameterizedByGradleVersion(lessThan = "9.0", stringValue = "A")
@ParameterizedByGradleVersion(lessThan = "8.0", stringValue = "B")
```

For version 7 do we give A or B?

File: docs/testing-guide.md (hunk 2 above)
Comment (FinlayRJW) [reply docs/testing-guide.md]: BEFOREeach

File: docs/testing-guide.md [hunk 9]
```diff
@@ -171,6 +172,45 @@ class CompatibilityTest {
...
 
+#### Version-Conditional Parameters
+
+Use `@ParameterizedByGradleVersion` to inject different String values based on the Gradle version under test.
+
+```java
+@Test
+@ParameterizedByGradleVersion(
+    name = "configOption",
+    otherwiseString = "newStyle",
```
CRogers: Can we put `otherwise` below when in all the examples and code etc? It reads wrong to have `otherwise` first.

FinlayRJW: fixed in [76b1cbc](https://github.com/palantir/gradle-plugin-testing/pull/358/commits/76b1cbc45e9c6d5a407cc1a4794739f6a282919a)

docs/testing-guide.md
```diff
@@ -171,6 +172,45 @@ class CompatibilityTest {
...
 The versions from `@AdditionallyRunWithGradle` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
 
+#### Version-Conditional Parameters
+
+Use `@ParameterizedByGradleVersion` to inject different String values based on the Gradle version under test.
+
+```java
+@Test
+@ParameterizedByGradleVersion(
+    name = "configOption",
```
CRogers: So this requires name to always be specified... I think I'm ok with it, I just think it adds some overhead in the basic case.

The other option would have been to be able to leave out the name in the basic case and have an annotation on the parameter:

```java
@Test
@ParameterizedByGradleVersion(
    when = @WhenVersion(lessThan = "8.0", stringValue = "legacyStyle"),
    otherwiseString = "newStyle")
void test(GradleInvoker gradle, RootProject project, @ParameterizedInject String configOption) {
```

I guess there isn't really a good name for `@ParameterizedInject` is the problem.

CRogers: The `@ParameterizedInject` makes it super obvious where it's coming from though.

FinlayRJW: would we also require `ParameterizedInject` if a name is given?

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/ParameterizedByGradleVersionBeforeEachFixtureTest.java
```diff
@@ -0,0 +1,56 @@
...
+import com.palantir.gradle.testing.execution.GradleInvoker;
+import com.palantir.gradle.testing.junit.GradlePluginTests;
+import com.palantir.gradle.testing.junit.ParameterizedByGradleVersion;
+import com.palantir.gradle.testing.junit.WhenVersion;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Fixture test for verifying @ParameterizedByGradleVersion works with @BeforeEach.
```
CRogers: I'd be tempted to group all these different fixture classes into a single file, and have nested classes.

docs/testing-guide.md
CRogers: I think it would be good for consistency at least.

FinlayRJW: fixed in [1634a5d](https://github.com/palantir/gradle-plugin-testing/pull/358/commits/1634a5daea7f8837c8bd8baaefc9a24cad1cbf20)

FinlayRJW: named `InjectByGradleVersion`

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/ParameterizedByGradleVersionBeforeEachFixtureTest.java
FinlayRJW: consolidated in [2b3b3dd](https://github.com/palantir/gradle-plugin-testing/pull/358/commits/2b3b3dd034ce91448cded3323bb9b1fdac1183bc)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WhenVersion.java
```diff
@@ -0,0 +1,31 @@
...
+package com.palantir.gradle.testing.junit;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/** Version condition for {@link ParameterizedByGradleVersion}. Matches when Gradle version is less than threshold. */
+@Target({})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface WhenVersion {
```
CRogers: I'd be tempted to put this as an inner class in `@ParameterizedByGradleVersion` - I don't think anything else will use.

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/ParameterizedByGradleVersionFixtureTest.java
```diff
@@ -0,0 +1,219 @@
...
+import com.palantir.gradle.testing.junit.ParameterizedByGradleVersion;
+import com.palantir.gradle.testing.junit.WhenVersion;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+public final class ParameterizedByGradleVersionFixtureTest {
+
+    @Nested
```
CRogers: _Technically_ I don't think we need these to be `@Nested` as we individually use these classes in the events based tests. Having them here kinda implies that the whole `ParameterizedByGradleVersionValuesTest` is being run as a test

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/InjectByGradleVersion.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.palantir.gradle.testing.junit;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/** Marks a String parameter to receive the value from {@link ParameterizedByGradleVersion}. */
```
felixdesouza: I don't think this doc is useful in that it's too specific. It will immediately go out of date the moment we add a new type. If anything, I would just leave it as `Marks a method parameter to receive the value from {@link ParameterizedByGradleVersion}`

docs/testing-guide.md
```diff
@@ -171,6 +172,62 @@ class CompatibilityTest {
...
+
+Use `@ParameterizedByGradleVersion` to inject different String values based on the Gradle version under test.
+
+```java
+@Test
+@ParameterizedByGradleVersion(
+    when = @WhenVersion(lessThan = "8.0", stringValue = "legacyStyle"),
+    otherwiseString = "newStyle")
+void test(GradleInvoker gradle, RootProject project, @InjectByGradleVersion String configOption) {
+    // configOption is "legacyStyle" for Gradle < 8.0, "newStyle" for Gradle >= 8.0
```
felixdesouza: nit: we should not put the ocmments explaining the different parameters here, at the very least, I don't think it should be a good example for the AI. Otherwise what's the point of this PR? If you can put the comment next to the `@WhenVersion` or even add a `description` or `why` field to `@WhenVersion` , that would serve as a better documentation for humans + AI than comments inside tbh.

If it's tautological, then maybe we eschew it entirely.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
```diff
@@ -0,0 +1,65 @@
...
+package com.palantir.gradle.testing.junit;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Injects a String parameter based on the Gradle version under test.
```
felixdesouza: I think we should remove any notion of type in this javadoc, it should be enough to say "Injects a parameter based on the Gradle version under test." For aforementioned reasons.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * <p>Simple usage (single parameter):
+ * <pre>{@code
+ * @ParameterizedByGradleVersion(
+ *     when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
+ *     otherwiseString = "new")
+ * void test(GradleInvoker invoker, RootProject project, @InjectByGradleVersion String behaviour) { }
+ * }</pre>
+ *
+ * <p>Multiple parameters (name required):
```
felixdesouza: Think we want to clarify that the name corresponds to the *parameter* name in the method.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
```diff
@@ -0,0 +1,65 @@
...
+ * <pre>{@code
+ * @ParameterizedByGradleVersion(
+ *     name = "style",
+ *     when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
+ *     otherwiseString = "new")
+ * @ParameterizedByGradleVersion(
+ *     name = "format",
+ *     when = @WhenVersion(lessThan = "9.0", stringValue = "classic"),
+ *     otherwiseString = "modern")
+ * void test(GradleInvoker invoker, @InjectByGradleVersion String style, @InjectByGradleVersion String format) { }
```
felixdesouza, line 46: A thought, if we have the name specified, then technically we don't really need `@InjectByGradleVersion` anymore? wdyt @CRogers @FinlayRJW ? you will if you use the single parameter version, but maybe you can make that required idk.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
```diff
@@ -0,0 +1,65 @@
...
+ * }</pre>
+ *
+ * <p>Conditions must be ordered by ascending version (lowest first).
```
felixdesouza: Should give an example of an bad annotation, and say it will fail when the tests run.

It would be nice if it failed to compile so you get quicker feedback, but that brings in an annotation processor, which is probably overkill.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -0,0 +1,115 @@
...
+    public static Optional<String> computeValue(Method method, String parameterName, GradleVersion gradleVersion) {
+        ParameterizedByGradleVersion[] annotations = method.getAnnotationsByType(ParameterizedByGradleVersion.class);
+
+        if (annotations.length == 0) {
+            return Optional.empty();
+        }
+
+        validate(annotations, method);
+
+        return findMatchingAnnotation(annotations, parameterName).map(anno -> {
```
felixdesouza: please call `anno` -> `annotation`, it doesn't really save anything and it's annoying to read

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
FinlayRJW: see https://github.com/palantir/gradle-plugin-testing/pull/358#discussion_r2742198229 - I think the idea is to keep it consistent even though its not strictly needed

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -0,0 +1,115 @@
...
+                .map(WhenVersion::stringValue)
+                .findFirst()
+                .orElseGet(annotation::otherwiseString);
+    }
+
+    private static void validate(ParameterizedByGradleVersion[] annotations, Method method) {
+        if (annotations.length > 1) {
+            validateAllHaveNames(annotations, method);
+        }
+        validateNoDuplicateNames(annotations, method);
```
felixdesouza: I think you can move this into the above? since you can't have duplicates with only one annotation

docs/testing-guide.md
```diff
@@ -171,6 +172,62 @@ class CompatibilityTest {
 
 The versions from `@AdditionallyRunWithGradle` are merged with the globally configured versions. When applied to both a class and a method, all versions are combined. Duplicate versions are automatically deduplicated.
 
+#### Version-Conditional Parameters
```
CRogers, line 175: Let's add a `beforeEach` example too.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -0,0 +1,115 @@
...
+            throw new IllegalStateException(
+                    "@ParameterizedByGradleVersion on %s.%s: name is required when multiple annotations are present"
+                            .formatted(method.getDeclaringClass().getSimpleName(), method.getName()));
+        }
+    }
+
+    private static void validateNoDuplicateNames(ParameterizedByGradleVersion[] annotations, Method method) {
+        List<String> nonEmptyNames = Arrays.stream(annotations)
+                .map(ParameterizedByGradleVersion::name)
+                .filter(name -> !name.isEmpty())
```
felixdesouza: don't think you need to filter here, the empty string isn't special in this instance, but won't necesasrily block on this

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -0,0 +1,115 @@
...
+                    .formatted(method.getDeclaringClass().getSimpleName(), method.getName()));
+        }
+    }
+
+    private static void validateOrdering(WhenVersion[] conditions, Method method) {
+        List<GradleVersion> versions = Arrays.stream(conditions)
+                .map(when -> new GradleVersion(when.lessThan()))
+                .toList();
+
+        boolean outOfOrder = IntStream.range(1, versions.size())
```
felixdesouza: nit: just use `!Comparators.isInOrder(versions, Comparator.naturalOrder())`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -0,0 +1,115 @@
...
+    private static void validateNoDuplicateNames(ParameterizedByGradleVersion[] annotations, Method method) {
+        List<String> nonEmptyNames = Arrays.stream(annotations)
+                .map(ParameterizedByGradleVersion::name)
+                .filter(name -> !name.isEmpty())
+                .toList();
+
+        Set<String> uniqueNames = Set.copyOf(nonEmptyNames);
+
+        if (uniqueNames.size() != nonEmptyNames.size()) {
+            throw new IllegalStateException("@ParameterizedByGradleVersion on %s.%s has duplicate name values"
```
felixdesouza: would be nice to show what is duplicated

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WhenVersion.java
felixdesouza: +1

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
```diff
@@ -0,0 +1,265 @@
...
+import org.junit.platform.engine.TestExecutionResult.Status;
+import org.junit.platform.engine.discovery.DiscoverySelectors;
+import org.junit.platform.testkit.engine.EngineExecutionResults;
+import org.junit.platform.testkit.engine.EngineTestKit;
+import org.junit.platform.testkit.engine.Event;
+
+final class ParameterizedByGradleVersionTest {
+
+    @Nested
+    class SingleParameter {
```
felixdesouza, line 36: funnily enough, this test should be a parameterised test lmao

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
```diff
@@ -0,0 +1,265 @@
...
+                            msg -> {
+                                assertThat(msg).contains("behavior=new");
+                                assertThat(msg).contains("GradleVersion: 8.5");
+                            });
+            assertThat(results.testEvents().skipped().count()).isZero();
+        }
+    }
+
+    @Nested
+    class WithDisabledConfigurationCache {
```
felixdesouza: why do we care about configuration cache here? i would have thought it entirely orthogonal

FinlayRJW: I wanted to check that the @DisabledConfigurationCache annotation was still working - previously when I used TestTemplate I broke interaction with @AddAdditionalGradleVersion - can probably remove now we are not using TestTemplate

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
```diff
@@ -0,0 +1,265 @@
...
+            Class<?> fixtureClass, String gradleVersion, String configurationCacheEnabled) {
+        return EngineTestKit.engine("junit-jupiter")
+                .selectors(DiscoverySelectors.selectClass(fixtureClass))
+                .configurationParameter("com.palantir.gradle.testing.gradle_versions_to_test", gradleVersion)
+                .configurationParameter(
+                        "com.palantir.gradle.testing.configuration_cache_enabled", configurationCacheEnabled)
+                .execute();
+    }
+
+    private static List<String> getExceptionMessages(List<Event> events) {
```
felixdesouza: can we use proper names please?

* `e` -> `event`
* `r` -> `result`

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
```diff
@@ -0,0 +1,265 @@
...
+        return events.stream()
+                .map(e -> e.getPayload(TestExecutionResult.class))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
```
felixdesouza: nit: `mapMulti(Optional::ifPresent)`

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
```diff
@@ -0,0 +1,265 @@
...
+            assertThat(results.testEvents().skipped().count()).isZero();
+        }
+    }
+
+    private static EngineExecutionResults runFixture(Class<?> fixtureClass, String gradleVersion) {
+        return runFixture(fixtureClass, gradleVersion, "false");
+    }
+
+    private static EngineExecutionResults runFixture(
+            Class<?> fixtureClass, String gradleVersion, String configurationCacheEnabled) {
```
felixdesouza: encoding a boolean as a string is not great, we should do the conversion where we actually push it through

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
```diff
@@ -0,0 +1,265 @@
...
+                assertThat(msg).contains("GradleVersion: 9.3.0");
+            });
+            assertThat(results.testEvents().skipped().count()).isZero();
+        }
+    }
+
+    @Nested
+    class SameNameAcrossMethods {
+        @Test
+        void same_name_allowed_on_before_each_and_test_for_old_gradle() {
```
felixdesouza, line 116: I don't think it matters to test multiple versions here. The thing you're verifying/asserting is that you can have the same name across two constructs. The version piece is orthogonal imo.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValuesTest.java
```diff
@@ -0,0 +1,255 @@
...
+
+            assertThatThrownBy(() ->
+                            ParameterizedByGradleVersionValues.computeValue(method, "first", new GradleVersion("8.0")))
+                    .isInstanceOf(IllegalStateException.class)
+                    .hasMessageContaining("name is required when multiple annotations are present");
+        }
+    }
+
+    // Test fixture classes with annotations for testing
+    static class ValidFixtures {
```
felixdesouza: can we move fixtures to the `testFixtures` project, for consistency sake please 🙏

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValuesTest.java
```diff
@@ -0,0 +1,255 @@
...
+                name = "first",
+                when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
+                otherwiseString = "new")
+        @ParameterizedByGradleVersion(
+                name = "second",
+                when = @WhenVersion(lessThan = "9.0", stringValue = "before 9"),
+                otherwiseString = "after 9")
+        void twoAnnotations() {}
+
+        void noAnnotation() {}
```
felixdesouza: none of these fixtures should be valid, you have no parameters in them, and they should fail

felixdesouza: for example, it is confusing that you specify `ParameterziedByGradleVersion` and there's nothing that it feeds into. Technically it's possible, but there is negative value with allowing this configuration.

felixdesouza: even if you consider it to not fail, at the very least have good examples that the AI can read from if it decides to look at these classes!

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValuesTest.java
```diff
@@ -0,0 +1,255 @@
...
+                    .isInstanceOf(IllegalStateException.class)
+                    .hasMessageContaining("name is required when multiple annotations are present");
+        }
+    }
+
+    // Test fixture classes with annotations for testing
+    static class ValidFixtures {
+
+        @ParameterizedByGradleVersion(otherwiseString = "default")
+        void noConditions() {}
```
felixdesouza: if there are no conditions, then this is pointless indirection. We should remove it.

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/ParameterizedByGradleVersionFixtureTest.java
felixdesouza: yes, let's remove it! These all end up being examples to the AI

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/WhenVersion.java
FinlayRJW: fix in 4c66f9e6b28f4ccb6a4ee6aa8d1de41ea1c52b7f

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/ParameterizedByGradleVersionFixtureTest.java
FinlayRJW: updated to be static in 8ed86454dda7ede98925f706944011dede780670 - without static or nested got

```
WARNING: TestEngine with ID 'junit-jupiter' encountered a non-critical issue during test discovery:

(1) [WARNING] Test class 'com.palantir.example.ParameterizedByGradleVersionFixtureTest$WithBeforeEach' must not be an inner class unless annotated with @Nested. It will not be executed.
    Source: ClassSource [className = 'com.palantir.example.ParameterizedByGradleVersionFixtureTest$WithBeforeEach', filePosition = null]
            at com.palantir.example.ParameterizedByGradleVersionFixtureTest$WithBeforeEach.<no-method>(SourceFile:0)
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/InjectByGradleVersion.java
FinlayRJW: fixed in 74a424424477dd150c5a6c3a20ac312ee7928710

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
FinlayRJW: fixed in 74a424424477dd150c5a6c3a20ac312ee7928710

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/ParameterizedByGradleVersionFixtureTest.java
```diff
@@ -0,0 +1,219 @@
...
+                @InjectByGradleVersion String format) {
+            rootProject.buildGradle().append("""
+                import org.gradle.util.GradleVersion
+                println "GradleVersion: ${GradleVersion.current().version}"
+                println "Style: %s"
+                println "Format: %s"
+                """, style, format);
+
+            String output = gradleInvoker.withArgs().buildsSuccessfully().output();
+            throw new RuntimeException("style=" + style + "|format=" + format + "|output=" + output);
```
felixdesouza: I kinda hate having to this sort of weird encoding and then having to parse out exception messages. We should just use `TestReporter`. You get a lot more structured formats e.g.

```java
@Nested
@GradlePluginTests
public class MultipleAnnotations {

    @Test
    @ParameterizedByGradleVersion(
            name = "style",
            when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
            otherwiseString = "new")
    @ParameterizedByGradleVersion(
            name = "format",
            when = @WhenVersion(lessThan = "9.0", stringValue = "classic"),
            otherwiseString = "modern")
    void test_with_two_parameters(
            GradleInvoker gradleInvoker,
            RootProject rootProject,
            TestReporter testReporter,
            @InjectByGradleVersion String style,
            @InjectByGradleVersion String format) {
            import org.gradle.util.GradleVersion
            println "GradleVersion: ${GradleVersion.current().version}"
            """, style, format);

        testReporter.publishEntry("style", style);
        testReporter.publishEntry("format", format);
    }
}
```

Here is how you consume it in tests:

```java
@Test
void multiple_parameters_receive_correct_values_for_old_gradle() {
    EngineExecutionResults results =
            runFixture(ParameterizedByGradleVersionFixtureTest.MultipleAnnotations.class, "7.6.4");

    results.testEvents()
            .assertStatistics(statistics -> statistics.skipped(0))
            .assertEventsMatchLooselyInOrder(
                    reportEntry(Map.of("style", "old")), reportEntry(Map.of("format", "classic")));
}
```

You need to statically import `EventConditions`.

You can also not bother having to do separate asserts for the statistics and just use the `assertStatistics` on the `Events` object. Junit has some pretty nice helpers in that regard, a shame to reimplement them imo.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
FinlayRJW: fixed in 0d80dd45a346ecee86be08430bce0e2ab5c47131

docs/testing-guide.md
FinlayRJW: fixed in be2e5b958bfba0316e17cf7962f53f5196a29eb2 also tautological is a great word!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
FinlayRJW: fixed in e71523c86f3ca2c61ec2c0dcff4a8da794a3e145

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: fixed in caf135595d3667558d1f549a42d8dafbf2093347

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: see caf135595d3667558d1f549a42d8dafbf2093347

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: fixed in caf135595d3667558d1f549a42d8dafbf2093347

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: nice! fixed in caf135595d3667558d1f549a42d8dafbf2093347

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: caf135595d3667558d1f549a42d8dafbf2093347 !!

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: whoops missed some - see efe1e1c44e9e6c689854d76dc7006bb396c7f2a3 and 05dda19e772c1a6aa51f3a72dc55f1192ab9fd24

docs/testing-guide.md
FinlayRJW: see 9bac140f76c8a3ec43c255124a489b5158b3f3b0

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
FinlayRJW: removed in 2bb421097d7d248d074edb46449c9d7539be7f63

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
FinlayRJW: fixed in 81fea593a243cd4971332f7a95762feb08dfaddd

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
FinlayRJW: fixed in 81fea593a243cd4971332f7a95762feb08dfaddd

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
FinlayRJW: resolved by removing CC test - 81fea593a243cd4971332f7a95762feb08dfaddd

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
FinlayRJW: very true - fixed in 3bd7fe8202f14b9bf35a26a54aa99f6a1ce4a26b

FinlayRJW: whoops deleted wrong test lol - fix in 2c873dabd14d8cc2aeb6cdefa7b379ffdb395d1c

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/ete/ParameterizedByGradleVersionTest.java
FinlayRJW: haha - yep fixed in a07d86afc8c38e3011aa7e9ccf2aec5529267a18

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValuesTest.java
FinlayRJW: I've added a new check `validateHasInjectParameter` to ensure that the the same number of injected params as params we want to inject, cleaned up the tests and add a new test for this in a6bc6d037451dfb708f20befba1b4afa5812b53a

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValuesTest.java
FinlayRJW: moved in dd6216fd66ccd81e03745061ccb87d75df5f7acd

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValuesTest.java
FinlayRJW: removed in 6187ac289296bddc15a1450af9c5a3f47c5a18cd

gradle-plugin-testing-junit/src/testFixtures/java/com/palantir/example/ParameterizedByGradleVersionFixtureTest.java
FinlayRJW: wow this is so much cleaner - b8ad19c686acc313ef1273cc729b5c1363e9ff8a

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
```diff
@@ -62,4 +70,14 @@
     WhenVersion[] when() default {};
 
     String otherwiseString();
+
+    /** Version condition. Matches when Gradle version is less than threshold. */
```
felixdesouza: Taken standalone, these semantics are confusing.

I don't think we've explained anywhere that the highest condition where `lessThan` is true wins. That should live somewhere, perhaps not on `WhenVersion`, I'm not sure yet, but maybe in `ParameterizedByGradleVersion#when`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -51,28 +53,46 @@ private static Optional<ParameterizedByGradleVersion> findMatchingAnnotation(
...
+                .map(ParameterizedByGradleVersion.WhenVersion::stringValue)
                 .findFirst()
                 .orElseGet(annotation::otherwiseString);
     }
 
     private static void validate(ParameterizedByGradleVersion[] annotations, Method method) {
+        validateHasInjectParameter(annotations, method);
         if (annotations.length > 1) {
             validateAllHaveNames(annotations, method);
+            validateNoDuplicateNames(annotations, method);
```
felixdesouza, line 77: I think you're missing one more assertion, because currently you have:
* all parameters need to have an inject parameter
* they're all named
* no duplicate names

however, there's nothing to prove that an annotation maps to a parameter when it should. I think you'll end up getting a bad looking error because it maps to `Optional#empty`

stepping back, this validation doesn't really need to happen on parameter resolution I guess as it's literally just looking at the methods and ensuring they're valid. It can happen when the test class is constructed, but it is a convenient place for it.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -51,28 +53,46 @@ private static Optional<ParameterizedByGradleVersion> findMatchingAnnotation(
...
+        }
+    }
+
+    private static void validateHasInjectParameter(ParameterizedByGradleVersion[] annotations, Method method) {
```
felixdesouza: this is not a great validation tbh, you could potentially annotate something else with `@InjectByGradleVersion` and it all adds up. 

I think a clearer and also more helpful message will tell you what is missing. This goes to my previous comment about matching up parameter names and the names in the annotation and throwing if they don't match.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValuesTest.java
```diff
@@ -155,101 +145,55 @@ void out_of_order_in_middle_throws() throws Exception {
...
+        }
+
+        @Test
+        void extra_inject_parameter_throws() throws Exception {
+            Method method = InvalidFixtures.class.getDeclaredMethod("extraInjectParameter", String.class, String.class);
+
+            assertThatThrownBy(() -> ParameterizedByGradleVersionValues.computeValue(
+                            method, "behaviour", new GradleVersion("8.0")))
+                    .isInstanceOf(IllegalStateException.class)
+                    .hasMessageContaining("requires a corresponding @InjectByGradleVersion parameter")
```
felixdesouza: this message doesn't make sense in this direction. The issue is that there's *fewer* `@ParameterizedByGradleVersion` parameters than injections, so by number alone (which is what you've been checking for) there is a corresponding `@InjectByGradleVersion`.

felixdesouza: it's like you want to validate in both directions i.e.
*  from `@ParameterizedByGradleVersion` is there a corresponding `@InjectByGradleVersion`.
* for each `@InjectByGradleVersion` is there a corresponding `@ParameterizedByGradleVersion`.

Both of those checks get you clearer error messages, and you also don't do any shenanigans with counting, but actually matching them up.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -82,28 +102,26 @@ private static void validateAllHaveNames(ParameterizedByGradleVersion[] annotati
...
 
     private static void validateNoDuplicateNames(ParameterizedByGradleVersion[] annotations, Method method) {
-        List<String> nonEmptyNames = Arrays.stream(annotations)
+        List<String> names = Arrays.stream(annotations)
                 .map(ParameterizedByGradleVersion::name)
-                .filter(name -> !name.isEmpty())
                 .toList();
 
-        Set<String> uniqueNames = Set.copyOf(nonEmptyNames);
+        Set<String> duplicates = names.stream()
```
felixdesouza: this is an strange way of making this change, it's effectively n^2.

let's move the `names.stream().filter(...)` out and replace it with a multiset, then you can query said multiset the same way.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersion.java
FinlayRJW: Good point - I'll add documentation to `when()` explaining the evaluation semantics (first matching condition wins, otherwiseString as fallback) and update the testing guide with the same explanation.

FinlayRJW: see [9a8df87](https://github.com/palantir/gradle-plugin-testing/pull/358/commits/9a8df877ce46c3ac2866db1612854259becfe657)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: resolved in 3fb172b9392874a6fd372fe199ac050021107072

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: yeah removed and switched over to `validateAnnotationsMatchParameters` now if we have more than one annotation we, check they all have unique names then we check each name matches a `@InjectByGradleVersion` and then check the other way around see 3fb172b9392874a6fd372fe199ac050021107072

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValuesTest.java
FinlayRJW: updated so now you should get messages like the following:

**1. Annotation name doesn't match any parameter name:**
  ```java
  @ParameterizedByGradleVersion(
      name = "wrongName",
      when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
      otherwiseString = "new")
  void test(@InjectByGradleVersion String param) {}
  ```
  **Error:** `@ParameterizedByGradleVersion on MyTest.test: no @InjectByGradleVersion parameter found for name(s): [wrongName]`

  ---

  **2. Parameter name doesn't match any annotation name:**
  ```java
  @ParameterizedByGradleVersion(
      name = "first",
      when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
      otherwiseString = "new")
  void test(@InjectByGradleVersion String first, @InjectByGradleVersion String extra) {}
  ```
  **Error:** `@ParameterizedByGradleVersion on MyTest.test: no annotation found for @InjectByGradleVersion parameter(s): [extra]`

  ---

  **3. Single unnamed annotation with multiple inject parameters:**
  ```java
  @ParameterizedByGradleVersion(
      when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
      otherwiseString = "new")
  void test(@InjectByGradleVersion String first, @InjectByGradleVersion String second) {}
  ```
  **Error:** `@ParameterizedByGradleVersion on MyTest.test without a name requires exactly one @InjectByGradleVersion parameter (found 2)`

  ---

  **4. Two annotations but only one has a matching parameter:**
  ```java
  @ParameterizedByGradleVersion(
      name = "first",
      when = @WhenVersion(lessThan = "8.0", stringValue = "old"),
      otherwiseString = "new")
  @ParameterizedByGradleVersion(
      name = "second",
      when = @WhenVersion(lessThan = "9.0", stringValue = "before"),
      otherwiseString = "after")
  void test(@InjectByGradleVersion String first) {}
  ```
  **Error:** `@ParameterizedByGradleVersion on MyTest.test: no @InjectByGradleVersion parameter found for name(s): [second]`

FinlayRJW: I think this should be pretty clear to a user what is going wrong now

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: fixed up the validation issues in [3fb172b](https://github.com/palantir/gradle-plugin-testing/pull/358/commits/3fb172b9392874a6fd372fe199ac050021107072) - in terms of moving - could use a `BeforeAllCallback`. But leaving it in parameter resolution for now since the error still surfaces before the test actually runs

docs/testing-guide.md
felixdesouza, line 197: add an example explanation for what is going to happen

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -116,6 +104,56 @@ private static void validateNoDuplicateNames(ParameterizedByGradleVersion[] anno
...
+                                        injectParameters.size()));
+            }
+            return;
+        }
+
+        // Validate: each annotation has a corresponding @InjectByGradleVersion parameter
+        Set<String> annotationNames = annotationsByName.keySet();
+        Set<String> missingParameters = annotationNames.stream()
+                .filter(name -> !injectParameterNames.contains(name))
+                .collect(Collectors.toSet());
```
felixdesouza: Can just use `Sets.difference` from guava

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
```diff
@@ -116,6 +104,56 @@ private static void validateNoDuplicateNames(ParameterizedByGradleVersion[] anno
...
+            throw new IllegalStateException(
+                    "@ParameterizedByGradleVersion on %s.%s: no @InjectByGradleVersion parameter found for name(s): %s"
+                            .formatted(
+                                    method.getDeclaringClass().getSimpleName(), method.getName(), missingParameters));
+        }
+
+        // Validate: each @InjectByGradleVersion parameter has a corresponding annotation
+        Set<String> missingAnnotations = injectParameterNames.stream()
+                .filter(name -> !annotationNames.contains(name))
+                .collect(Collectors.toSet());
```
felixdesouza: `Sets.difference` as well here pls

FinlayRJW: switched here a06849fb68fa4e9fdb47d68bee8ec3c0494f535f

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ParameterizedByGradleVersionValues.java
FinlayRJW: fixed here a06849fb68fa4e9fdb47d68bee8ec3c0494f535f

docs/testing-guide.md
FinlayRJW: fixed in 32d6f64df8554b09873d801948706e38c34b3038

CRogers (approved):

PR #361, opened 2026-01-30
Add best practices section to testing guide

CRogers:
## Before this PR

The testing guide and migration instructions don't document common best practices for test setup and assertions.

## After this PR

Adds a "Best Practices" section to the testing guide and a "Test Setup Best Practices" section to the migration instructions, covering:

- **Combine multiple append() calls**: Use a single `append()` with one text block
- **No leading newlines**: Start text block content immediately after `"""`
- **Group related setup**: Separate unrelated setup calls with blank lines
- **Skip unnecessary setup**: Don't call `rootProjectName()` unless needed
- **Prefer exact assertions**: Use `isEqualTo()` with text blocks instead of `contains()`
- **Test edge cases**: Test with Gradle properties explicitly unset

These patterns were learned from code review feedback and should help produce cleaner, more maintainable tests.

## Possible downsides?

None - documentation only.


docs/junit-migration-excavator-instructions.md
```diff
@@ -178,6 +178,43 @@ assertThat(result).task(":myProject:checkstyleMain").succeeded();
...
+    group 'com.example'
+    version '1.0.0'
+    """);
+
+// WRONG - multiple appends
+project.buildGradle().append("group 'com.example'");
+project.buildGradle().append("version '1.0.0'");
+```
+
+**No leading newlines**: Start text block content immediately after `"""`:
```
CRogers: This is because it's more important for the _test file_ to be readable than the actual generated code in the test, which is unlikely to ever be read

docs/junit-migration-excavator-instructions.md
```diff
@@ -178,6 +178,43 @@ assertThat(result).task(":myProject:checkstyleMain").succeeded();
...
+// WRONG - unnecessary leading newline
+project.buildGradle().append("""
+
+    myPlugin {
+    """);
+```
+
+**Skip unnecessary setup**: Don't call `settingsGradle().rootProjectName()` unless you need a custom name.
+
+**Use exact assertions**: Prefer `isEqualTo()` with text blocks over `contains()` for file content - makes expected output obvious.
```
CRogers: For lock file content only

FinlayRJW:
👍

PR #364, opened 2026-02-02
Add  `GradleInvoker.with(Options)`

crogoz:
## After this PR

Allows extending the arguments passed to the gradleInvoker with extra options (initial thread [here](https://github.com/palantir/gradle-plugin-testing/pull/344#discussion_r2716677620)) eg: 

```
gradle.with(Options.builder()
    .args(...)
    .testingEnvironmentVariable(...)
    .somethingElse(...)
    .builder()).buildsSuccessfully()
```
`.withArgs` falls back to using `.with`

==COMMIT_MSG==
Add  `GradleInvoker.with(Options)`
==COMMIT_MSG==


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/Options.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.palantir.gradle.testing.execution;
+
+import com.palantir.gradle.testing.ImmutablesStyle;
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@ImmutablesStyle
+public interface Options {
```
CRogers: I'd make this an abstract class as it makes it a bit easier for us to hide and change stuff in future if we want.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/Options.java
```diff
@@ -0,0 +1,38 @@
...
+    List<String> args();
+
+    class Builder extends ImmutableOptions.Builder {}
+
+    static Builder builder() {
+        return new Builder();
+    }
+
+    static Builder from(Options options) {
+        return builder().args(options.args());
```
CRogers: Immutables already make a method on the `Builder` to do this - you can just do `builder().from(otherInstance)`. I think it would be nice to make an instance override for this so you don't need a static method - something like:

```
default Builder asBuilder() {
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -17,26 +17,29 @@
...
-import com.google.errorprone.annotations.RestrictedApi;
-import com.palantir.gradle.testing.RestrictedCreation;
 import java.nio.file.Path;
-import java.util.Arrays;
 import org.gradle.testkit.runner.GradleRunner;
 
-public record DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) implements GradleInvoker {
-    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
-    public DefaultGradleInvoker {}
+public final class DefaultGradleInvoker extends GradleInvoker {
```
felixdesouza: do we want this to be `public`?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -17,26 +17,29 @@
...
+
+    private final Path rootProjectDir;
+    private final GradleVersion gradleVersion;
+
+    public DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) {
```
felixdesouza: think we still want the restricted api on this if it is public? don't think there's any need to ever construct this or am I missing something

gradle-test-versions-config/src/test/java/com/palantir/gradle/plugintesting/GradleTestVersionsConfigTest.java
```diff
@@ -25,7 +25,7 @@
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
-public class GradleTestVersionsConfigTest {
+public final class GradleTestVersionsConfigTest {
```
felixdesouza, line 28: didn't realise that you'd have to make this final :S

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/Options.java
crogoz: do you mean allowing for `someOptions.asBuilder().addAllArgs(args)`?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
crogoz: It is required by [MavenRepoPublisher](https://github.com/palantir/gradle-plugin-testing/blob/develop/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java#L39) which is in a different package

felixdesouza: it's a little weird that `MavenRepoPublisher` needs to have instantiate it directly. We should just pass it through. I'm not even sure whether we'd want to run all of the decorators around it? probably not, but in any case, it does feel pretty weird. If push came to shove, we could do `GradleInvoker.create` inside `MavenRepoPublisher` for example

crogoz: I can change it in a different PR (since this PR maintains the existing `public class` behavior). I also don't think we need to use all the decorators around the gradleInvoker used here.

crogoz: opened https://github.com/palantir/gradle-plugin-testing/pull/370. let me know what you think

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -18,16 +18,21 @@
 
 import java.lang.management.ManagementFactory;
 import java.nio.file.Path;
+import java.util.Arrays;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public interface GradleInvoker {
+public abstract class GradleInvoker {
```
CRogers: I meant make the `Options` an abstract class (it makes the immutables classes a bit easier to change in the future), `GradleInvoker` should stay an interface as it will be implemented by others.

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/GradleInvoker.java
```diff
@@ -18,16 +18,21 @@
...
 
-    Logger log = LoggerFactory.getLogger(GradleInvoker.class);
+    private static Logger log = LoggerFactory.getLogger(GradleInvoker.class);
```
CRogers: Presumably this logger will disappear when we do the decorator thing anyway.

CRogers (approved):

PR #370, opened 2026-02-03
Package-private `DefaultGradleInvoker` class

crogoz:
## Before this PR

`MavenRepoPublisher` needs to instantiate a defaultInvoker (it shouldn't require the extra decorations eg. Configuration Cache etc.) (discussion [here](https://github.com/palantir/gradle-plugin-testing/pull/364#discussion_r2758525463))

## After this PR

Made package-private `DefaultGradleInvoker` class. Introduced restricted internal method for creating a defaultInvoker.
==COMMIT_MSG==
Package-private `DefaultGradleInvoker` class
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
@@ -17,7 +17,6 @@
 package com.palantir.gradle.testing.maven;
 
 import com.palantir.gradle.testing.execution.GradleInvoker;
-import com.palantir.gradle.testing.execution.GradleVersion;
 import com.palantir.gradle.testing.project.RootProject;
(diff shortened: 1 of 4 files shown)


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/maven/MavenRepoPublisher.java
```diff
@@ -36,7 +35,7 @@ final class MavenRepoPublisher {
 
     MavenRepoPublisher(Path projectRoot, Path mavenRepoUrl, GradleVersion gradleVersion) {
```
felixdesouza: can we pull it all the way up? for example, `GradleVersion` only ever exists to make a `GradleInvoker`, and it's a little weird to pass in your dependency's dependencies. Similarly passing in a `RootProject`. 

In fairness this can merge as in, I can push something up as a FLUP, do not want to block on the main PRs

felixdesouza (approved):

PR #378, opened 2026-02-11
[PropertiesFile] SetProperty adds/updates property value

crogoz:
## After this PR

Required by https://github.com/palantir/gradle-jdks/pull/661/changes#r2712867634. `setProperty` sets the property to the given value, replacing the existing value if the key already exists, or appending a new property line if it does not.

==COMMIT_MSG==
[PropertiesFile] appendProperty de-dups & checks values
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/properties/PropertiesFile.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/properties/PropertiesFile.java
@@ -30,7 +30,11 @@ public record PropertiesFile(Path path) implements ProjectFile<PropertiesFile> {
     public PropertiesFile {}
 
     public PropertiesFile appendProperty(String key, String value) {
-        return appendLine("%s=%s", key, value);
+        String propertyValue = String.format("%s=%s", key, value);
+        if (!text().contains(propertyValue)) {
+            return appendLine("%s=%s", key, value);
+        }
+        return this;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/properties/PropertiesFile.java
```diff
@@ -30,7 +33,23 @@ public record PropertiesFile(Path path) implements ProjectFile<PropertiesFile> {
...
+        String text = Files.exists(path()) ? text() : "";
+        Matcher matcher = getPropertyMatcher(text, key);
+        if (!matcher.find()) {
+            return appendLine("%s=%s", key, value);
+        }
+        String existingValue = matcher.group(1);
+        if (!existingValue.equals(value)) {
+            throw new IllegalArgumentException(
+                    String.format("Property '%s' already exists with a different value: '%s'", key, existingValue));
+        }
```
CRogers: What happens if we want to change the property though? I think there are use cases for this, ie in a test you change the value from som,e setup. I guess doing this in `appendProperty` might be a bit odd - probably should have called it `setProperty`

crogoz: I've modified it to not throw and always update the value (added the javadoc as well) [here](https://github.com/palantir/gradle-plugin-testing/pull/378/commits/248ff2609868257f05c8d99cb9eaa5a3092284ca)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/properties/PropertiesFile.java
```diff
@@ -21,16 +21,44 @@
...
-        return appendLine("%s=%s", key, value);
+        return setProperty(key, value);
+    }
+
+    /**
+     * Sets the property {@code key} to the given {@code value}, replacing the existing value if the key already exists,
+     * or appending a new property line if it does not.
+     */
+    public PropertiesFile setProperty(String key, String value) {
+        String originalText = Files.exists(path()) ? text() : "";
```
CRogers: Can't we write this all inside of `edit`? Rather than reading the text twice?

crogoz: addressed in https://github.com/palantir/gradle-plugin-testing/pull/378/commits/c3a4f8e045776d4ee6d0732cd564033e02f8e47b

CRogers (approved):

PR #393, opened 2026-02-24
Create `InvocationResult` from `UnexpectedBuildResultException`

crogoz:
## After this PR

==COMMIT_MSG==
Create `InvocationResult` from `UnexpectedBuildResultException`
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResult.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResult.java
@@ -19,6 +19,7 @@ package com.palantir.gradle.testing.execution;
 import com.palantir.gradle.testing.assertion.InvocationResultAssert;
 import java.util.Optional;
 import org.gradle.testkit.runner.BuildResult;
+import org.gradle.testkit.runner.UnexpectedBuildResultException;
 
 public final class InvocationResult {
     private final BuildResult buildResult;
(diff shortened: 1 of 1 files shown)


CRogers (approved):

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/InvocationResult.java
```diff
@@ -35,6 +36,10 @@ public Optional<TaskResult> task(String taskPath) {
         return Optional.ofNullable(buildResult.task(taskPath)).map(TaskResult::new);
     }
 
+    public static InvocationResult from(UnexpectedBuildResultException result) {
```
CRogers, line 39: Actually, on second thoughts I don't think this is ideal. Really, we should avoid exposing test kit types at all here. I think a better approach would be to execute a task without expecting success or failure. There's a method for this on `GradleRunner`: https://docs.gradle.org/current/javadoc/org/gradle/testkit/runner/GradleRunner.html#run()

I wonder if we could expose that on `GradleInvoker`? It's a bit of a break, because it's a new method `GradleInvoker` implementations need to handle, but I think we can deal with it just fine? Perhaps we could make `buildsSuccessfully` and `buildsWithFailure` `default` methods that just call `run`? Then

CRogers: Actually, I don't think this is useable as `BuildResult` does not tell you whether it succeeded or failed, so is not useful for this use case.

PR #400, opened 2026-02-25
Discover test classes from all source sets, not just `test`

FinlayRJW:
## Before this PR

`DiscoverTestClassesTask` was only configured with the `test` source set's classpath and source files. Tests in other source sets like `integrationTest` were never discovered by tasks like `discoverNebulaTestClassesToMigrate` or `discoverGradlePluginTestsWithDisabledConfigurationCache`.

## After this PR

`addTestClassesDiscoveryTasks` now iterates all non-main source sets via `sourceSetContainer.all()`, so discovery tasks find tests in `integrationTest` and any other custom source sets. Using `all()` also handles source sets added by plugins that apply after ours.

## Possible downsides?

Source sets unrelated to testing will have their files included in the discovery scan. This shouldn't cause issues since those source sets won't contain nebula test classes or `@DisabledConfigurationCache` annotations — they'll just add files to the classpath without affecting discovery results.

## Test plan

- New test `nebula_tests_in_non_test_source_sets_are_discovered` creates an `integrationTest` source set with a nebula test class and verifies it's discovered
- Existing tests continue to pass

### DIFF
--- a/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
+++ b/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
@@ -201,9 +201,12 @@ public abstract class PluginTestingPlugin implements Plugin<Project> {
 
         project.getTasks().withType(DiscoverTestClassesTask.class, task -> {
             SourceSetContainer sourceSetContainer = project.getExtensions().getByType(SourceSetContainer.class);
-            SourceSet testSourceSet = sourceSetContainer.getByName(SourceSet.TEST_SOURCE_SET_NAME);
-            task.getTestClasspath().setFrom(testSourceSet.getRuntimeClasspath());
-            task.getTestSourceFiles().setFrom(testSourceSet.getAllSource());
+            sourceSetContainer.all(sourceSet -> {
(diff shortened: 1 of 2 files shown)


gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -201,9 +201,12 @@ private static void addTestClassesDiscoveryTasks(Project project) {
 
         project.getTasks().withType(DiscoverTestClassesTask.class, task -> {
             SourceSetContainer sourceSetContainer = project.getExtensions().getByType(SourceSetContainer.class);
-            SourceSet testSourceSet = sourceSetContainer.getByName(SourceSet.TEST_SOURCE_SET_NAME);
-            task.getTestClasspath().setFrom(testSourceSet.getRuntimeClasspath());
-            task.getTestSourceFiles().setFrom(testSourceSet.getAllSource());
+            sourceSetContainer.all(sourceSet -> {
```
felixdesouza, line 204: I'm a little bit skeptical about the `.all` vs `.configureEach` but I think there's possibility for it to be weird with the nested `tasks().withType()` hmm, can think about it for another time

felixdesouza (approved):

PR #404, opened 2026-02-27
[docs] Migration docs update

crogoz:
## After this PR

==COMMIT_MSG==
[docs] Migration docs update
==COMMIT_MSG==

### DIFF
--- a/docs/junit-migration-excavator-instructions.md
+++ b/docs/junit-migration-excavator-instructions.md
@@ -67,6 +67,7 @@ These comments should be copied over to the new test files as they are created t
 
 # General Instructions
 - You MUST migrate ALL tests from the original file. Do not skip any tests or only migrate a "representative sample". Every single test method in the original Groovy file must have a corresponding test method in the new Java file.
+- Always add both `@GradlePluginTests` and `@DisabledConfigurationCache` to the migrated test classes.
 - Test names should be changed to a snake_case_english_sentence in all lower case.
 - Keep the comments from the original Groovy tests when writing the new tests.
 - Method name mappings from old to new framework:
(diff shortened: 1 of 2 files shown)


docs/junit-migration-excavator-instructions.md
```diff
@@ -201,21 +202,6 @@ rootProject.buildGradle().append("""
 rootProject.buildGradle().plugins().add("com.palantir.baseline");
 ```
 
-## Always Provide a Reason for @DisabledConfigurationCache
```
crogoz, line 204: Unfortunately the LLM won't know the reason for `@DisabledConfigurationCache` unless it runs the tests & finds out some error messages. 

By default we will be adding `@DisabledConfigurationCache` and there is an extra excavator that will try to remove these annotations.

docs/testing-guide.md
```diff
@@ -498,12 +498,12 @@ void create_java_class(RootProject project) {
 
 ### Properties Files
 
-Manage `gradle.properties` and other properties files.  Use `appendProperty(key, value)` for properties files, not `append()`
+To add key-value pairs to properties files such as `gradle.properties` or custom files like `versions.props`, use the `setProperty(key, value)` method:
```
crogoz, line 501: `appendProperty` is deprecated, we should use `setProperty` instead.

FinlayRJW:
👍

docs/junit-migration-excavator-instructions.md
felixdesouza: I think if that's the case, we should say that it's going to be disabled for "rollout" and something else will come by and enable it. Because then you're back to the beginning of "why is configuration cache disabled for this test". Currently, there's no way to tell if something was added manually, or added automatically, and whether it should be removed, or whether it can never work. This was the main reason behind making this docs change.

If you want to add a section to not try to fix configuration cache errors, that's also an option but I don't think we should be doing this.

PR #410, opened 2026-03-06
Add `customGradleUserHome` in `Options`

crogoz:
## Before this PR

Default `gradleUserHome` directory set [here](https://github.com/gradle/gradle/blob/master/platforms/extensibility/test-kit/src/main/java/org/gradle/testkit/runner/internal/DefaultGradleRunner.java#L85), shared by all tests (`<project>/build/tmp/test/work/.gradle-test-kit`)

## After this PR

Allowing GradlePluginTests to override the `gradleUserHome`. Required by gradle-jdks: https://github.com/palantir/gradle-jdks/pull/635/changes#diff-51800d21937c5a5e7382c93a573dfa6d4e33e4f2a1f2cde9f96e107ed8cfd003R42

==COMMIT_MSG==
Add `customGradleUserHome` in `Options`
==COMMIT_MSG==

### DIFF
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
@@ -42,6 +42,7 @@ record DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) im
                                 .map(e -> String.format("-P__TESTING_%s=%s", e.getKey(), e.getValue()))
                                 .toList())
                         .build());
+        options.customGradleUserHome().ifPresent(gradleUserHome -> runner.withTestKitDir(gradleUserHome.toFile()));
 
         return new DefaultGradleInvocation(runner);
     }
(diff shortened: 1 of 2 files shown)


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -42,6 +42,7 @@ public GradleInvocation with(Options options) {
                                 .map(e -> String.format("-P__TESTING_%s=%s", e.getKey(), e.getValue()))
                                 .toList())
                         .build());
+        options.customGradleUserHome().ifPresent(gradleUserHome -> runner.withTestKitDir(gradleUserHome.toFile()));
```
crogoz, line 45: does it make sense to override the default value for testkitdir to the `<gradle-plugin-testing>/<TestClass>`  ?

felixdesouza: I don’t think so, I think there might be perf implications if so, how many things are we downloading per user home etc.

felixdesouza (approved):

PR #413, opened 2026-03-11
Add `IncludedBuild` support for composite build testing

FinlayRJW:
## Before this PR
With us starting to support included builds, and some tests already testing included builds, it felt that the framework was missing native support for example see https://github.com/palantir/gradle-consistent-versions/pull/1508#discussion_r2917321124

Users testing plugins that interact with `includeBuild` directives had to manually create directories, write `settings.gradle` files, and wire up `includeBuild` directives — verbose and error-prone.

## After this PR

Adds `IncludedBuild` as a new injectable parameter type. It works the same way as `SubProject` — the parameter name becomes the build name, and everything is wired up automatically.

**Parameter injection:**
```java
@Test
void composite_build(GradleInvoker gradle, IncludedBuild myLib) {
    myLib.buildGradle().plugins().add("java-library");
    gradle.withArgs("tasks").buildsSuccessfully();
}
```

**Programmatic creation:**
```java
IncludedBuild lib = rootProject.includedBuild("my-lib");
```

## Test plan

- `included_build_uses_exact_parameter_name` — parameter name becomes build name
- `included_build_manually` — programmatic creation via `rootProject.includedBuild()`
- `included_build_with_subproject` — subprojects within an included build
- `included_build_settings_gradle` — custom settings.gradle content is evaluated
- `included_build_gradle_properties` — gradle.properties are readable from the included build
- `nested_included_build` — included build within an included build
- `can_request_the_same_included_build_multiple_times` — idempotency

### DIFF
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
@@ -72,14 +72,19 @@ public final class ConfigurationCacheInvocation implements GradleInvocation {
         // check if there is a configuration time error that isn't related to configuration cache errors or any script
         // exceptions
         if (result.output().contains("org.gradle.api.ProjectConfigurationException:")
-                || result.output().contains("org.gradle.api.GradleScriptException:")
-                || result.output().contains("which is the same as a project of the main build")) {
+                || result.output().contains("org.gradle.api.GradleScriptException:")) {
             return result;
         }
 
         assertConfigCacheStored(result);
-        InvocationResult configurationCacheResult = secondGradleInvocation.buildsSuccessfully();
-        assertConfigCacheReused(configurationCacheResult);
+        try {
+            InvocationResult configurationCacheResult = secondGradleInvocation.buildsSuccessfully();
+            assertConfigCacheReused(configurationCacheResult);
+        } catch (UnexpectedBuildFailure unexpectedBuildFailure) {
+            if (unexpectedBuildFailure.getBuildResult().getOutput().contains("configuration cache cannot be reused")) {
+                throwReuseFailure(new InvocationResult(unexpectedBuildFailure.getBuildResult()));
+            }
+        }
         return result;
     }
 
@@ -112,19 +117,23 @@ public final class ConfigurationCacheInvocation implements GradleInvocation {
 
     private void assertConfigCacheReused(InvocationResult result) {
         if (!result.output().contains("Configuration cache entry reused.")) {
-            throw new UnexpectedConfigurationCacheFailure(String.format("""
-                Configuration cache reuse failure: The second run failed to reuse the cached configuration.
-
-                This test runs with configuration cache enabled (`gradleTestUtils.configurationCacheEnabled=true`).
-                Test sequence:
-                    ✓ First run: Successfully executed tasks and stored configuration cache
-                    ✗ Second run: Failed to reuse configuration cache during dry-run verification
-
-                Please check the output below for specific configuration cache problems.
-
-                Output:
-                %s
-                """, result.output()), result);
+            throwReuseFailure(result);
         }
+    }
+
+    private static void throwReuseFailure(InvocationResult result) {
+        throw new UnexpectedConfigurationCacheFailure(String.format("""
+            Configuration cache reuse failure: The second run failed to reuse the cached configuration.
+
+            This test runs with configuration cache enabled (`gradleTestUtils.configurationCacheEnabled=true`).
+            Test sequence:
+                ✓ First run: Successfully executed tasks and stored configuration cache
+                ✗ Second run: Failed to reuse configuration cache during dry-run verification
+
+            Please check the output below for specific configuration cache problems.
+
+            Output:
+            %s
+            """, result.output()), result);
     }
 }
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
@@ -72,7 +72,8 @@ public final class ConfigurationCacheInvocation implements GradleInvocation {
         // check if there is a configuration time error that isn't related to configuration cache errors or any script
         // exceptions
         if (result.output().contains("org.gradle.api.ProjectConfigurationException:")
-                || result.output().contains("org.gradle.api.GradleScriptException:")) {
+                || result.output().contains("org.gradle.api.GradleScriptException:")
+                || result.output().contains("which is the same as a project of the main build")) {
             return result;
         }
 
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
@@ -81,7 +81,9 @@ public final class ConfigurationCacheInvocation implements GradleInvocation {
             InvocationResult configurationCacheResult = secondGradleInvocation.buildsSuccessfully();
             assertConfigCacheReused(configurationCacheResult);
         } catch (UnexpectedBuildFailure unexpectedBuildFailure) {
-            assertConfigCacheReused(new InvocationResult(unexpectedBuildFailure.getBuildResult()));
+            if (unexpectedBuildFailure.getBuildResult().getOutput().contains("configuration cache cannot be reused")) {
+                throwReuseFailure(new InvocationResult(unexpectedBuildFailure.getBuildResult()));
+            }
         }
         return result;
     }
@@ -115,19 +117,23 @@ public final class ConfigurationCacheInvocation implements GradleInvocation {
 
     private void assertConfigCacheReused(InvocationResult result) {
         if (!result.output().contains("Configuration cache entry reused.")) {
-            throw new UnexpectedConfigurationCacheFailure(String.format("""
-                Configuration cache reuse failure: The second run failed to reuse the cached configuration.
-
-                This test runs with configuration cache enabled (`gradleTestUtils.configurationCacheEnabled=true`).
-                Test sequence:
-                    ✓ First run: Successfully executed tasks and stored configuration cache
-                    ✗ Second run: Failed to reuse configuration cache during dry-run verification
-
-                Please check the output below for specific configuration cache problems.
-
-                Output:
-                %s
-                """, result.output()), result);
+            throwReuseFailure(result);
         }
+    }
+
+    private static void throwReuseFailure(InvocationResult result) {
+        throw new UnexpectedConfigurationCacheFailure(String.format("""
+            Configuration cache reuse failure: The second run failed to reuse the cached configuration.
+
+            This test runs with configuration cache enabled (`gradleTestUtils.configurationCacheEnabled=true`).
+            Test sequence:
+                ✓ First run: Successfully executed tasks and stored configuration cache
+                ✗ Second run: Failed to reuse configuration cache during dry-run verification
+
+            Please check the output below for specific configuration cache problems.
+
+            Output:
+            %s
+            """, result.output()), result);
     }
 }
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
@@ -81,9 +81,7 @@ public final class ConfigurationCacheInvocation implements GradleInvocation {
             InvocationResult configurationCacheResult = secondGradleInvocation.buildsSuccessfully();
             assertConfigCacheReused(configurationCacheResult);
         } catch (UnexpectedBuildFailure unexpectedBuildFailure) {
-            if (unexpectedBuildFailure.getBuildResult().getOutput().contains("configuration cache cannot be reused")) {
-                throwReuseFailure(new InvocationResult(unexpectedBuildFailure.getBuildResult()));
-            }
+            assertConfigCacheReused(new InvocationResult(unexpectedBuildFailure.getBuildResult()));
         }
         return result;
     }
@@ -117,23 +115,19 @@ public final class ConfigurationCacheInvocation implements GradleInvocation {
 
     private void assertConfigCacheReused(InvocationResult result) {
         if (!result.output().contains("Configuration cache entry reused.")) {
-            throwReuseFailure(result);
+            throw new UnexpectedConfigurationCacheFailure(String.format("""
+                Configuration cache reuse failure: The second run failed to reuse the cached configuration.
+
+                This test runs with configuration cache enabled (`gradleTestUtils.configurationCacheEnabled=true`).
+                Test sequence:
+                    ✓ First run: Successfully executed tasks and stored configuration cache
+                    ✗ Second run: Failed to reuse configuration cache during dry-run verification
+
+                Please check the output below for specific configuration cache problems.
+
+                Output:
+                %s
+                """, result.output()), result);
         }
-    }
-
-    private static void throwReuseFailure(InvocationResult result) {
-        throw new UnexpectedConfigurationCacheFailure(String.format("""
-            Configuration cache reuse failure: The second run failed to reuse the cached configuration.
-
-            This test runs with configuration cache enabled (`gradleTestUtils.configurationCacheEnabled=true`).
-            Test sequence:
-                ✓ First run: Successfully executed tasks and stored configuration cache
-                ✗ Second run: Failed to reuse configuration cache during dry-run verification
-
-            Please check the output below for specific configuration cache problems.
-
-            Output:
-            %s
-            """, result.output()), result);
     }
 }
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
@@ -48,20 +48,11 @@ public record SettingsGradleFile(Path path) implements GradleFile {
     }
 
     public SettingsGradleFile includeBuild(String buildPath) {
-        checkNoConflict(buildPath, "include '%s'".formatted(buildPath), "a subproject");
         return idempotentAppend("includeBuild '%s'".formatted(buildPath));
     }
 
     public SettingsGradleFile include(String projectPath) {
-        checkNoConflict(projectPath, "includeBuild '%s'".formatted(projectPath), "an included build");
         return idempotentAppend("include '%s'".formatted(projectPath));
-    }
-
-    private void checkNoConflict(String name, String conflictingLine, String conflictingType) {
-        if (Files.exists(path) && text().contains(conflictingLine)) {
-            throw new IllegalArgumentException(
-                    "'%s' is already registered as %s and cannot also be used here".formatted(name, conflictingType));
-        }
     }
 
     private SettingsGradleFile idempotentAppend(@Language("Gradle") String line) {
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
@@ -47,15 +47,20 @@ public record SettingsGradleFile(Path path) implements GradleFile {
         return this;
     }
 
-    public SettingsGradleFile include(String projectPath) {
-        @Language("Gradle")
-        String includeLine = "include '%s'".formatted(projectPath);
+    public SettingsGradleFile includeBuild(String buildPath) {
+        return idempotentAppend("includeBuild '%s'".formatted(buildPath));
+    }
 
-        if (Files.exists(path) && text().contains(includeLine)) {
+    public SettingsGradleFile include(String projectPath) {
+        return idempotentAppend("include '%s'".formatted(projectPath));
+    }
+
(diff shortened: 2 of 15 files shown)


gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -122,6 +123,95 @@ void can_request_the_same_subproject_multiple_times_without_issue(GradleInvoker
...
+        outerLib.buildGradle().append("""
+            println "outer build name: ${name}"
+            """);
+
+        gradle.withArgs().buildsSuccessfully().assertThat().output().contains("outer build name: outerLib");
+        gradle.withArgs().buildsSuccessfully().assertThat().output().contains("inner build name: inner");
+    }
+
+    @Test
+    void can_request_the_same_included_build_multiple_times(GradleInvoker gradle, RootProject rootProject) {
```
crogoz, line 213: what happens if we have name clashes between a an includedBuild and a subProject ? would they use the same path under the hood ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
```diff
@@ -52,6 +43,33 @@ default SubProject subproject(String name) {
...
+        includedBuildRoot.gradlePropertiesFile().setProperty("org.gradle.parallel", "true");
+
+        rootProject().settingsGradle().includeBuild(name);
+
+        return new IncludedBuild(includedBuildDir, includedBuildRoot);
+    }
+
+    private static Path createValidatedDirectory(Path parent, String name, String type) {
+        Preconditions.checkArgument(!name.contains(":"), "%s names must not %s contain colons", type, name);
+        Preconditions.checkArgument(!name.contains("/"), "%s names must not %s contain slashes", type, name);
```
crogoz: nit: I think the message is a bit weird, probably it should something look like: 

```suggestion
        Preconditions.checkArgument(!name.contains(":"), "%s names must not contain colons (project name was %s)", type, name);
        Preconditions.checkArgument(!name.contains("/"), "%s names must not contain slashes (project name was %s)", type, name);
```

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
FinlayRJW: I think they would share the same path and then I expect gradle we complain - will add a check to ensure that a user can't have a subproject and a included build with the same name

FinlayRJW: added a guard check in [9e15ef0](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/9e15ef07f20e18fa39a93b8d527bf3797b8d32d8)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
```diff
@@ -52,6 +43,35 @@ default SubProject subproject(String name) {
         return new SubProject(subprojectDir, rootProject());
     }
 
+    default IncludedBuild includedBuild(String name) {
+        Path includedBuildDir = createValidatedDirectory(rootProject().path(), name, "Included build");
+
+        RootProject includedBuildRoot = new RootProject(includedBuildDir);
+        includedBuildRoot.settingsGradle().rootProjectName(name);
+        includedBuildRoot.gradlePropertiesFile().setProperty("org.gradle.parallel", "true");
```
crogoz: I wonder if this logic shouldn't actually go inside the `IncludedBuild` constructor 
where the constructor would get a name & path and the class would contain a private RootProject member (you might need to convert it to a normal class not a record)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
```diff
@@ -47,15 +47,29 @@ public SettingsGradleFile rootProjectName(String rootProjectName) {
         return this;
     }
 
+    public SettingsGradleFile includeBuild(String buildPath) {
+        checkNoConflict(buildPath, "include '%s'".formatted(buildPath), "a subproject");
```
crogoz: hmm, it feels weird that the includeBuild would know about how subprojects are added, I think it is fine to not have these checks and just rely on gradle's error message.

FinlayRJW: check removed in [c8d2d3f](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/c8d2d3fc67f542698cbd430480536e53e70429c3)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
FinlayRJW: yeah I think that makes sense incase we ever create a new included build from elsewhere - fixed in [47bbbc5](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/47bbbc56c4f7316b17ae25552af875f70e0b41d3)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/files/gradle/SettingsGradleFile.java
FinlayRJW: nice gradle gives a clear error message:

```
Included build in /Volumes/git/gradle-plugin-testing/gradle-plugin-testing-junit/build/gradle-plugin-testing/ProjectUsagesTest/included build with subproject/8.14.3/myLib has name 'myLib' which is the same as a project of the main build.
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/IncludedBuild.java
```diff
@@ -0,0 +1,61 @@
...
+    public IncludedBuild(String name, Path path) {
+        this.path = path;
+        this.rootProject = new RootProject(path);
+        this.rootProject.settingsGradle().rootProjectName(name);
+        this.rootProject.gradlePropertiesFile().setProperty("org.gradle.parallel", "true");
+    }
+
+    @Override
+    public Path path() {
+        return path;
```
crogoz: I don't think you need to keep path as a member, you can infer it from rootProject.getPath()

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -213,24 +212,6 @@ void can_request_the_same_included_build_multiple_times(GradleInvoker gradle, Ro
         gradle.withArgs().buildsSuccessfully().assertThat().output().contains("included build name: shared");
     }
 
-    @Test
-    void subproject_then_included_build_with_same_name_throws(RootProject rootProject) {
```
crogoz: I'd keep a test that has both a subproject & an included build with the same name & expect a buildFailure when you run the gradleInvoker.

FinlayRJW: added back in [8954b7c](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/8954b7cd3e20b2f17ebdbd47f58faddcf3fa4fb0)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/IncludedBuild.java
FinlayRJW: simplified in [8954b7c](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/8954b7cd3e20b2f17ebdbd47f58faddcf3fa4fb0)

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -212,6 +214,16 @@ void can_request_the_same_included_build_multiple_times(GradleInvoker gradle, Ro
         gradle.withArgs().buildsSuccessfully().assertThat().output().contains("included build name: shared");
     }
 
+    @Test
+    void subproject_then_included_build_with_same_name_throws(GradleInvoker gradle, RootProject rootProject) {
+        rootProject.subproject("shared");
+        rootProject.includedBuild("shared");
+
+        assertThatThrownBy(() -> gradle.withArgs().buildsSuccessfully())
```
crogoz: why not `gradle.withArgs().buildsWithFailure()` ?

FinlayRJW: I did this as the user would expect the build to pass right? But then it happens to fail - if I run with `buildsWithFailure` I get an error with `ConfigurationCacheInvocation#buildWithFailure` from `InvocationResult configurationCacheResult = secondGradleInvocation.buildsSuccessfully();`

I would need to add `|| result.output().contains("org.gradle.api.GradleException:")` to `check if there is a configuration time error that isn't related to configuration cache errors or any script exceptions` - I wasn't sure if `GradleException` fell into `isn't related to configuration cache errors or any script exceptions`

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
```diff
@@ -52,6 +43,31 @@ default SubProject subproject(String name) {
         return new SubProject(subprojectDir, rootProject());
     }
 
+    default IncludedBuild includedBuild(String name) {
```
CRogers: It feels strange to have this on `GradleProject`. Logically, included builds are at the `Settings` level. This allows you to make included builds on subprojects which seems misleading. Could we not just have t his method on the `SettingsFile` object? Or at least on `RootProject` rather than `GradleProject`?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/IncludedBuild.java
```diff
@@ -0,0 +1,59 @@
...
+import java.nio.file.Path;
+
+/**
+ * When injected as a parameter in JUnit test methods, the parameter name will be used as the build name exactly.
+ * For example, {@code IncludedBuild myLib} creates an included build named "myLib".
+ * <br>
+ * When injected, the included build will be registered via {@code includeBuild} in the
+ * root project's settings.gradle.
+ */
+public final class IncludedBuild implements GradleProject {
```
CRogers: So this `IncludedBuild` is logically a `RootProject`. I'd be tempted to keep it this way by making `RootProject` an interface, then having `IncludedBuild` implement it, and what currently returns `new RootProject` should instead return a `new TopLevelRootProject` or similar (return type would still be `RootProject` though.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
```diff
@@ -122,6 +125,105 @@ void can_request_the_same_subproject_multiple_times_without_issue(GradleInvoker
...
+    void included_build_settings_gradle(GradleInvoker gradle, IncludedBuild myLib) {
+        myLib.settingsGradle().append("""
+            println "included settings evaluated"
+            """);
+        myLib.buildGradle().append("""
+            println "included build name: ${name}"
+            """);
+
+        gradle.withArgs().buildsSuccessfully().assertThat().output().contains("included settings evaluated");
+        gradle.withArgs().buildsSuccessfully().assertThat().output().contains("included build name: myLib");
```
CRogers: Can we avoid running these twice? Can't you just do multiple `.contains` on the output assertj matcher?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/IncludedBuild.java
FinlayRJW: changed over in [13dd1e5](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/13dd1e57e30b6a656c9674eb2da9c0da61864d9f)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
FinlayRJW: moved to root project in [c9b5708](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/c9b570811b00068f4ab37fda225bd101371fb8b2)

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
FinlayRJW: fixed in [e4a6a7e](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/e4a6a7e5e5eabae96d1698fc1d5510f6a82f24d0)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/IncludedBuild.java
```diff
@@ -0,0 +1,44 @@
...
+ * When injected, the included build will be registered via {@code includeBuild} in the
+ * root project's settings.gradle.
+ */
+public final class IncludedBuild implements RootProject {
+    private final Path path;
+
+    @RestrictedApi(explanation = RestrictedCreation.EXPLANATION, allowedOnPath = RestrictedCreation.ALLOWED_ON_PATH)
+    public IncludedBuild(String name, Path path) {
+        this.path = path;
+        settingsGradle().rootProjectName(name);
```
CRogers: I wonder if we should call this `${INCLUDED_BUILD_NAME}-root` to be like most of our real repos?

I'm not actually sure `settingsGradle().rootProjectName(name);` changes much as by default it will use the directory name, which is `name`.

gradle-plugin-testing-junit/src/test/java/com/palantir/gradle/testing/integration/ProjectUsagesTest.java
FinlayRJW: I think the safest option is to just match for `which is the same as a project of the main build` in `ConfigurationCacheInvocation#buildsWithFailure` see [0f2dd4c](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/0f2dd4cd6f7cae52831bf702f9d2a84b705e167e)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/IncludedBuild.java
FinlayRJW: remove the `settingsGradle().rootProjectName(name)` entirely in [d18cd32](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/d18cd323603a51aaba2805a9feedbb75a633626e)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
```diff
@@ -72,7 +72,8 @@ public InvocationResult buildsWithFailure() {
         // check if there is a configuration time error that isn't related to configuration cache errors or any script
         // exceptions
         if (result.output().contains("org.gradle.api.ProjectConfigurationException:")
-                || result.output().contains("org.gradle.api.GradleScriptException:")) {
+                || result.output().contains("org.gradle.api.GradleScriptException:")
+                || result.output().contains("which is the same as a project of the main build")) {
```
crogoz: hmm, I think it would be better to change the seconfInvocation below. there might be other GradleExceptions (eg. https://github.com/gradle/gradle/blob/8c5fe1fd7bf5acb8feabbd4531784975ea063e02/subprojects/composite-builds/src/main/java/org/gradle/composite/internal/DefaultIncludedBuildRegistry.java#L213) that might throw and we would end with the same `UnexpectedBuildFailure` error - something like: [here](https://github.com/palantir/gradle-plugin-testing/commit/c79661edaa90ff55d82ab4524ed0784cf286a72c)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/RootProject.java
```diff
@@ -16,30 +16,48 @@
...
+        Preconditions.checkArgument(
+                !name.contains("/"), "Included build names must not contain slashes (project name was %s)", name);
+
+        Path includedBuildDir = path().resolve(name);
+
+        try {
+            Files.createDirectories(includedBuildDir);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
```
crogoz: could we reuse this logic for checking the name & creating the directories in both GradleProject#subproject & RootProject#includedBuild ?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
FinlayRJW: okay I've switched to 

```java
try {
    InvocationResult configurationCacheResult = secondGradleInvocation.buildsSuccessfully();
    assertConfigCacheReused(configurationCacheResult);
} catch (UnexpectedBuildFailure unexpectedBuildFailure) {
    if (unexpectedBuildFailure.getBuildResult().getOutput().contains("configuration cache cannot be reused")) {
        throwReuseFailure(new InvocationResult(unexpectedBuildFailure.getBuildResult()));
    }
}
```

Which I think should be a catch-all for configuration cache errors but not other `GradleExceptions`

FinlayRJW: Is this what you meant?

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/RootProject.java
FinlayRJW: good idea fixed in - [34f6641](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/34f6641394c34350d454bcbecbc3da9224f18c03)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
crogoz: yes, that looks great

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/project/GradleProject.java
```diff
@@ -52,6 +43,23 @@ default SubProject subproject(String name) {
         return new SubProject(subprojectDir, rootProject());
     }
 
+    static Path createChildProjectDir(Path parentDir, String name) {
```
crogoz: I think you could have this as a `default Path createChildProjectDir(String name) {` and the parentDir will always be `path()`

FinlayRJW: [0451e71](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/0451e717c8fc2fdd96baa7f4e76c766cf9d1beda)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/ConfigurationCacheInvocation.java
```diff
@@ -77,8 +77,14 @@ public InvocationResult buildsWithFailure() {
...
 
         assertConfigCacheStored(result);
-        InvocationResult configurationCacheResult = secondGradleInvocation.buildsSuccessfully();
-        assertConfigCacheReused(configurationCacheResult);
+        try {
+            InvocationResult configurationCacheResult = secondGradleInvocation.buildsSuccessfully();
+            assertConfigCacheReused(configurationCacheResult);
+        } catch (UnexpectedBuildFailure unexpectedBuildFailure) {
+            if (unexpectedBuildFailure.getBuildResult().getOutput().contains("configuration cache cannot be reused")) {
+                throwReuseFailure(new InvocationResult(unexpectedBuildFailure.getBuildResult()));
```
crogoz, line 85: maybe we can add a test for this case as well ? I think this one should cover this case: 

```
    @Test
    void fails_when_configuration_cache_cannot_be_reused(GradleInvoker invoker, RootProject rootProject) {
                tasks.register("some") {
                    def propsFile = project.layout.projectDirectory.file('gradle.properties').asFile
                    doLast {
                        def props = new Properties()
                        propsFile.withInputStream { props.load(it) }
                        props.setProperty('version', 'two')
                        propsFile.withOutputStream { props.store(it, null) }
                        throw new RuntimeException("Version changed from one to two")
                    }
                }
            """);

        rootProject.gradlePropertiesFile().setProperty("version", "one");
        assertThatThrownBy(() -> invoker.withArgs("some").buildsWithFailure())
                .isInstanceOf(UnexpectedConfigurationCacheFailure.class)
                .hasMessageContaining(
                        "Configuration cache reuse failure: The second run failed to reuse the cached configuration.");
    }
```

FinlayRJW: added in [4c568dd](https://github.com/palantir/gradle-plugin-testing/pull/413/commits/4c568ddaa7d733044adf5f53a6695c84d8309014)

crogoz (approved):

PR #439, opened 2026-04-15
Replace `ProjectLayout.getSettingsDirectory()` with Gradle 8.8 compatible API when running gradle less than 8.10

FinlayRJW:
##  Before this PR
`ProjectLayout.getSettingsDirectory()` was introduced in Gradle 8.10. Consumers still on Gradle 8.8 get a `NoSuchMethodError` when applying this plugin.                                                                                                                                                                  

## After this PR                                                                                                                                                                                                                                                                                                         
Uses `project.getRootProject().getLayout().getProjectDirectory()` to resolve the `gradle/gradle-test-versions.yml` path when running on gradle version less than 8.10

## Possible downsides?
This assumes the settings directory is the same as the root project directory, which is true for standard single-root-project builds. Although I think this is fine for all of our current usages

### DIFF
--- a/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
+++ b/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
@@ -116,6 +116,14 @@ public abstract class PluginTestingPlugin implements Plugin<Project> {
         Provider<String> testDependenciesFileAbsolutePath = project.provider(() ->
                 testDependencyVersions.get().getOutputFile().get().getAsFile().getAbsolutePath());
 
+        RegularFile testVersionsConfigPath;
+        if (GradleVersion.current().compareTo(GradleVersion.version("8.10")) >= 0) {
+            testVersionsConfigPath = getProjectLayout().getSettingsDirectory().file("gradle/gradle-test-versions.yml");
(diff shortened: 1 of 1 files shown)


gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -116,6 +112,9 @@ private void doApply(Project project) {
         Provider<String> testDependenciesFileAbsolutePath = project.provider(() ->
                 testDependencyVersions.get().getOutputFile().get().getAsFile().getAbsolutePath());
 
+        RegularFile testVersionsConfigPath =
+                project.getRootProject().getLayout().getProjectDirectory().file("gradle/gradle-test-versions.yml");
```
kelvinou01: hmm this might make our efforts to support included builds more more difficult. how many repos are being blocked on 8.8? how much friction is there to bump them to 8.14.3, then roll out the framework?

FinlayRJW: I've put in c2540ca7e8d82a5ef86e9ae806cff85688c14202 so this will only effect older versions

kelvinou01 (approved):

PR #444, opened 2026-04-20
Add `gradleDistributionBaseUrl` extension property

FinlayRJW:
## Before this PR

Tests always download Gradle distributions from `services.gradle.org`. There was no way to configure an alternative distribution server (e.g. a local mirror or cache).

## After this PR

Adds a `gradleDistributionBaseUrl` property to the `gradleTestUtils` extension. When set, tests download Gradle from `{baseUrl}/gradle-{version}-bin.zip` instead of the default server.

```groovy
gradleTestUtils {
    gradleDistributionBaseUrl = 'https://my-mirror.example.com/gradle-distributions'
}
```

The value flows through as a system property (`com.palantir.gradle.testing.gradle_distribution_base_url`) to the test JVM, where `DefaultGradleInvoker` uses `GradleRunner.withGradleDistribution(URI)` instead of `withGradleVersion(String)`. When unset, behavior is unchanged.

## Testing

- New integration test in `PluginTestingJunitPluginTest` verifies the system property is correctly propagated from the extension to the inner test JVM

### DIFF
--- a/gradle-plugin-testing-junit/build.gradle
+++ b/gradle-plugin-testing-junit/build.gradle
@@ -37,3 +37,8 @@ dependencies {
 gradleTestUtils {
     configurationCacheEnabled = true
 }
+
+// TODO(gradle-plugin-testing): remove once buildscript classpath is updated to a version that sets this property
+tasks.named('test', Test) {
+    systemProperty 'com.palantir.gradle.testing.gradle_distribution_base_url', 'https://services.gradle.org/distributions'
+}
--- a/gradle-plugin-testing-junit/build.gradle
+++ b/gradle-plugin-testing-junit/build.gradle
@@ -22,6 +22,7 @@ dependencies {
 
     implementation 'commons-io:commons-io'
     implementation 'com.google.guava:guava'
+    implementation project(':plugin-testing-core')
 
     compileOnly 'org.immutables:value'
     annotationProcessor 'org.immutables:value'
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
@@ -18,6 +18,7 @@ package com.palantir.gradle.testing.execution;
 
 import com.google.common.collect.ImmutableList;
 import com.google.errorprone.annotations.RestrictedApi;
+import com.palantir.gradle.plugintesting.GradleDistributionBaseUrl;
 import com.palantir.gradle.testing.RestrictedCreation;
 import java.io.ByteArrayOutputStream;
 import java.io.OutputStreamWriter;
@@ -61,8 +62,8 @@ record DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) im
 
     private URI gradleDistributionUri() {
         String baseUrl = Optional.ofNullable(
-                        System.getProperty("com.palantir.gradle.testing.gradle_distribution_base_url"))
-                .orElse("https://services.gradle.org/distributions");
+                        System.getProperty(GradleDistributionBaseUrl.GRADLE_DISTRIBUTION_BASE_URL_SYSTEM_PROPERTY))
+                .orElse(GradleDistributionBaseUrl.DEFAULT_BASE_URL);
         return URI.create(baseUrl + "/gradle-" + gradleVersion.version() + "-bin.zip");
     }
 
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
@@ -27,7 +27,6 @@ import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.List;
-import java.util.Optional;
 import org.gradle.testkit.runner.GradleRunner;
 
 record DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) implements GradleInvoker {
@@ -61,9 +60,7 @@ record DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) im
     }
 
     private URI gradleDistributionUri() {
-        String baseUrl = Optional.ofNullable(
-                        System.getProperty(GradleDistributionBaseUrl.GRADLE_DISTRIBUTION_BASE_URL_SYSTEM_PROPERTY))
-                .orElse(GradleDistributionBaseUrl.DEFAULT_BASE_URL);
+        String baseUrl = System.getProperty(GradleDistributionBaseUrl.GRADLE_DISTRIBUTION_BASE_URL_SYSTEM_PROPERTY);
         return URI.create(baseUrl + "/gradle-" + gradleVersion.version() + "-bin.zip");
     }
 
--- a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
+++ b/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
@@ -26,7 +26,6 @@ import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.List;
-import java.util.Optional;
 import org.gradle.testkit.runner.GradleRunner;
 
 record DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) implements GradleInvoker {
@@ -60,9 +59,7 @@ record DefaultGradleInvoker(Path rootProjectDir, GradleVersion gradleVersion) im
     }
 
     private URI gradleDistributionUri() {
-        String baseUrl = Optional.ofNullable(
-                        System.getProperty("com.palantir.gradle.testing.gradle_distribution_base_url"))
-                .orElse("https://services.gradle.org/distributions");
+        String baseUrl = System.getProperty("com.palantir.gradle.testing.gradle_distribution_base_url");
         return URI.create(baseUrl + "/gradle-" + gradleVersion.version() + "-bin.zip");
     }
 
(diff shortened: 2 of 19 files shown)


gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -57,6 +59,13 @@ public GradleInvocation with(Options options) {
         return new DefaultGradleInvocation(runner, title, capturedOutput);
     }
 
+    private URI gradleDistributionUri() {
+        String baseUrl = Optional.ofNullable(
+                        System.getProperty("com.palantir.gradle.testing.gradle_distribution_base_url"))
```
felixdesouza: is there a reason to not use the constant you just defined?

FinlayRJW: I don't think `plugin-testing-core` is accessible from `gradle-plugin-testing-junit` I followed the pattern we used for `ConfigurationCacheHelper`

FinlayRJW: I don't see a reason that `gradle-plugin-testing-junit` shouldn't have a dependency on `plugin-testing-core` though - will add

FinlayRJW: fixed in [e22dec6](https://github.com/palantir/gradle-plugin-testing/pull/444/commits/e22dec65fce231f0f87af032e5458ae4d8fdc487)

gradle-plugin-testing-junit/build.gradle
```diff
@@ -22,6 +22,7 @@ dependencies {
 
     implementation 'commons-io:commons-io'
     implementation 'com.google.guava:guava'
+    implementation project(':plugin-testing-core')
```
FinlayRJW, line 25: allow us to access `GradleDistributionBaseUrl` from core

plugin-testing-core/src/main/java/com/palantir/gradle/plugintesting/ConfigurationCacheHelper.java
```diff
@@ -17,5 +17,5 @@
 package com.palantir.gradle.plugintesting;
 
 public record ConfigurationCacheHelper() {
-    static final String CONFIG_CACHE_SYSTEM_PROPERTY = "com.palantir.gradle.testing.configuration_cache_enabled";
+    public static final String CONFIG_CACHE_SYSTEM_PROPERTY = "com.palantir.gradle.testing.configuration_cache_enabled";
```
FinlayRJW, line 20: can remove from this PR as not strictly needed but figured we may as well reuse this now we have added 
```groovy
implementation project(':plugin-testing-core')
```

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -57,6 +60,13 @@ public GradleInvocation with(Options options) {
         return new DefaultGradleInvocation(runner, title, capturedOutput);
     }
 
+    private URI gradleDistributionUri() {
+        String baseUrl = Optional.ofNullable(
+                        System.getProperty(GradleDistributionBaseUrl.GRADLE_DISTRIBUTION_BASE_URL_SYSTEM_PROPERTY))
+                .orElse(GradleDistributionBaseUrl.DEFAULT_BASE_URL);
```
crogoz: I think the default value should already be set [here](https://github.com/palantir/gradle-plugin-testing/pull/444/changes#diff-011b68e8b94dff0d445c7d76f30fd9d0bd53f06236a65e8ed7ad3dd2e34495b8R53)

FinlayRJW: I think we need the orElse as the `gradle-plugin-testing-junit` module's tests compile `DefaultGradleInvoker` from local source (new code that reads the system property), but the `PluginTestingPlugin` that  configures the test task comes from the buildscript classpath - can either remove in a FLUP or remove the convention

crogoz: I think we can add the systemProp in the build.gradle and remove it afterwards (this is what we had to do for CC as well):

``` 
tasks.named('test', Test) {
 // TODO... remove this
 systemProperty 'GRADLE_DISTRIBUTION_BASE_URL_SYSTEM_PROPERTY', '...'
}
```

FinlayRJW: fab fixed in [68541b5](https://github.com/palantir/gradle-plugin-testing/pull/444/commits/68541b5b3cc6f4b4fc11a3c8336bbd98beade7b2)

FinlayRJW: and [04f961c](https://github.com/palantir/gradle-plugin-testing/pull/444/commits/04f961c9a25a98c7e424c1faec7b3a574326e7e3)

FinlayRJW: Will remove in a FLUP straight after

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/execution/DefaultGradleInvoker.java
```diff
@@ -57,6 +59,11 @@ public GradleInvocation with(Options options) {
         return new DefaultGradleInvocation(runner, title, capturedOutput);
     }
 
+    private URI gradleDistributionUri() {
```
crogoz, line 62: I realized now, that you can pass this directly the value for GRADLE_DISTRIBUTION_BASE_URL_SYSTEM_PROPERTY from the `extension.getConfigurationParameter` - same as we do when we check for CC if it is enabled or not - https://github.com/palantir/gradle-plugin-testing/blob/0df3f009eed6d4ee53ded631a653fe9f1814494a/gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/ConfigurationCacheDecorator.java#L38

FinlayRJW: good idea! have switched over to `getConfigurationParameter` still need the `build.gradle` workaround for some tests i.e. `InvocationResultUsagesTest` as these don't run through `EngineTestKit.engine(...)` - they're direct `@Test` methods via `@GradlePluginTests`, so there's no `.configurationParameter(...)`

see 3c57e60f613494e3f32f8ffc47e8c60316bf7b10

gradle-plugin-testing-junit/build.gradle
felixdesouza: let's double check we're not bringing in gradle dependencies

FinlayRJW: the core project doesnt declare `gradleApi()` or `gradleTestKit()` deps and a `./gradlew :plugin-testing-core:dependencies  | grep -E "org\.gradle|gradle-api|gradle-test-kit"` returns nothing - so I'm pretty sure its okay

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/MavenRepoStore.java
```diff
@@ -46,7 +47,7 @@ private static MavenRepo initializeMavenRepo(ExtensionContext context) {
 
         clearDirectory(repositoryDirectory);
 
-        return new MavenRepo(repositoryDirectory, gradleVersion);
+        return new MavenRepo(repositoryDirectory, gradleVersion, GradleInvoker.readGradleDistributionBaseUrl(context));
```
crogoz: what do you think about moving some parts of MavenRepo to the logic here ? This way we'd have all the GradleInvoker method invocation in the initialization, mavenRepo wouldn't need to care how the create the invoker and what params to send ? 

```
Path repoDirPath = repositoryDirectory.resolve("localMavenRepositoryPublisherProject");
Path mavenRepoUrl = repositoryDirectory.resolve("localMavenRepository");
String gradleDistributionBaseUrl = GradleInvoker.readGradleDistributionBaseUrl(context);
GradleInvoker invoker = GradleInvoker.getInternalDefaultInvoker(repoDirPath, gradleVersion, gradleDistributionBaseUrl)
return new MavenRepo(mavenRepoUrl, new TopLevelRootProject(repoDirPath), invoker);
```

and the constructor would look like: 
```
public MavenRepo(Path mavenRepoUrl, RootProject rootProject, GradleInvoker invoker) {
```

FinlayRJW: yeah that makes sense, updated in [b3a893b](https://github.com/palantir/gradle-plugin-testing/pull/444/commits/b3a893b12b5a3b78802218a6d4f3b423d6e584b8)

gradle-plugin-testing-junit/src/main/java/com/palantir/gradle/testing/junit/MavenRepoStore.java
```diff
@@ -46,7 +48,12 @@ private static MavenRepo initializeMavenRepo(ExtensionContext context) {
 
         clearDirectory(repositoryDirectory);
 
-        return new MavenRepo(repositoryDirectory, gradleVersion);
+        Path repoDirPath = repositoryDirectory.resolve("localMavenRepositoryPublisherProject");
+        Path mavenRepoUrl = repositoryDirectory.resolve("localMavenRepository").toAbsolutePath();
```
crogoz: nit:
```suggestion
        Path mavenRepoUrl = repositoryDirectory.resolve("localMavenRepository");
```

crogoz (dismissed):

crogoz:
👍

PR #449, opened 2026-04-23
Isolate GCV types behind helper class so plugin loads without GCV on the classpath

FinlayRJW:
## Before this PR                                                                                                                                                                                                                                                                                                     

`PluginTestingPlugin` imported `VersionsLockExtension` directly and referenced it inside a `withPlugin("com.palantir.consistent-versions", ...)` callback. Even though that callback only fires when GCV is applied, the type reference sits on `PluginTestingPlugin` itself,                  

```
║ > Failed to apply plugin class 'com.palantir.gradle.plugintesting.PluginTestingPlugin'.                                                                                                             
║    > Could not create plugin of type 'PluginTestingPlugin'.                                                                                                                                         
║       > Could not generate a decorated class for type PluginTestingPlugin.                                                                                                                          
║          > com/palantir/gradle/versions/VersionsLockExtension$ScopeConfigurer   
```                                                                                            

## After this PR                                                                                                                                                                                                                                                                                                      

The GCV call is moved into a new package-private `GradleConsistentVersionsInterop` helper. `PluginTestingPlugin` no longer imports any GCV types

### DIFF
--- a/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleConsistentVersionsInterop.java
+++ b/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleConsistentVersionsInterop.java
@@ -1,0 +1,34 @@
+/*
+ * (c) Copyright 2026 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/GradleConsistentVersionsInterop.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.palantir.gradle.plugintesting;
+
+import com.palantir.gradle.versions.VersionsLockExtension;
+import org.gradle.api.Project;
+
+/**
+ * Isolates GCV types so {@link PluginTestingPlugin} can be decorated without GCV on the classpath. Only call from
```
CRogers: Decorated?

FinlayRJW: fixed in [ebe5313](https://github.com/palantir/gradle-plugin-testing/pull/449/commits/ebe5313b7cb83d29ee9d44e46fc1a8c517210f1f)

CRogers (approved):

PR #460, opened 2026-05-20
Add `discoverGradlePluginTestsAnnotatedTests` task

FinlayRJW:
## Before this PR
There is no built-in way to ask "does this project have any `@GradlePluginTests`-annotated test classes?". The existing `discoverGradlePluginTestsWithDisabledConfigurationCache` task only finds the strict subset that also has `@DisabledConfigurationCache`.

## After this PR
A new `discoverGradlePluginTestsAnnotatedTests` task is registered on every project that applies `com.palantir.gradle-plugin-testing`. It finds any class annotated with `@GradlePluginTests` regardless of whether `@DisabledConfigurationCache` is also present.

==COMMIT_MSG==
Register discoverGradlePluginTestsAnnotatedTests task to discover any test class annotated with @GradlePluginTests
==COMMIT_MSG==

## Possible downsides?

### DIFF
--- a/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
+++ b/gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
@@ -202,6 +202,23 @@ public abstract class PluginTestingPlugin implements Plugin<Project> {
                                             "com.palantir.gradle.testing.junit.DisabledConfigurationCache"));
                         });
 
+        project.getTasks().register("discoverGradlePluginTestsAnnotatedTests", DiscoverTestClassesTask.class, task -> {
+            task.getTestClassType().set("java");
+            // Override the default output paths so this task does not collide with
+            // discoverGradlePluginTestsWithDisabledConfigurationCache (both use testClassType=java).
+            task.getDiscoveredTestsFile()
+                    .set(project.getLayout()
+                            .getBuildDirectory()
+                            .file("tests-discovery/gradle-plugin-tests-annotated/test-classes"));
+            task.getOutputFile()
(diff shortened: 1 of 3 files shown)


gradle-plugin-testing/src/main/java/com/palantir/gradle/plugintesting/PluginTestingPlugin.java
```diff
@@ -202,6 +202,23 @@ private static void addTestClassesDiscoveryTasks(Project project) {
...
+            // Override the default output paths so this task does not collide with
+            // discoverGradlePluginTestsWithDisabledConfigurationCache (both use testClassType=java).
+            task.getDiscoveredTestsFile()
+                    .set(project.getLayout()
+                            .getBuildDirectory()
+                            .file("tests-discovery/gradle-plugin-tests-annotated/test-classes"));
+            task.getOutputFile()
+                    .set(project.getLayout()
+                            .getBuildDirectory()
+                            .file("tests-discovery/gradle-plugin-tests-annotated/test-classes-paths"));
```
crogoz: maybe let's change the `DiscoverTestClassesTask`'s outputRootDirectory to `tests-discovery/<task_name>/<test-type/>` instead. The only usage that breaks is https://pl.ntr/2B3 but I can fix this in a FLUP

FinlayRJW: and https://pl.ntr/2B4  right?

FinlayRJW: updated in a27ec6284a52f17fe7ca1de44744ca1784d81d6f

crogoz (approved):

PR #475, opened 2026-06-17
Move `-P__TESTING` property from hardcoded default to annotation-based decorator

crogoz:
## Before this PR
<!-- What's wrong with the current state of the world and why change it now? -->
Plugins like `gradle-utils:environment-variables` have an `envVarOrFromTestingProperty` method that branches on `__TESTING`  when true it reads from Gradle properties, when false it reads from actual environment variables. With `__TESTING=true` hardcoded  after user args, there was no way to test the real environment variable path since                                                                                                                                                                                                                              
  the last `-P` flag wins in Gradle. 

## After this PR
 - Replaces the hardcoded -P__TESTING=true in DefaultGradleInvoker with a `TestingPropertyDecorator` registered on `@GradlePluginTests`                                                                                                                                           
 - Adds `@DisabledTestingProperty` annotation to opt out of the testing property on a class or method                                                                                                   
 - Makes `@RegistersGradleInvokerDecorator` repeatable so` @GradlePluginTests` can register multiple decorators, and updates the decorator discovery to use `getAnnotationsByType` accordingly                                                                                                                                                                                                                                                                             

<!-- User-facing outcomes this PR delivers go below -->
==COMMIT_MSG==
Move -P__TESTING property from hardcoded default to annotation-based decorator
==COMMIT_MSG==

## Possible downsides?
<!-- Please describe any way users could be negatively affected by this PR. -->


felixdesouza:
a thought, wouldn't this be a good example for a decorator? not opposed to this PR, but largely curious what you think here.

crogoz:
> a thought, wouldn't this be a good example for a decorator? not opposed to this PR, but largely curious what you think here.

Didn't think about it, but yes, switched now to using annotations

felixdesouza:
👍

